PR #218, opened 2013-09-30
fix load more arrivals button to always show results

bbodenmiller:
* If no arrivals in next 35 minutes on view load tells user:
photo 3
* Load more arrivals button will automatically search in 30 minute increments until it finds new arrivals or it has searched 1 day in the future
* If no results for 1 day tells users
    * If there was no results when load more button was pressed:
photo 4
    * If there was already results when load more button was pressed:
photo 5
* Adds loading animation for load more arrivals button
photo 1
* Adds loading animation for load more arrivals button when refresh is occurring (rather than just disabling)
photo 2
* Removes loading bar when searching through 30 minute increments for next arrival since we don't know how many 30 minute incruments we will need to look through
capture
* Changes load more button click color to blue so as to match other buttons
* Reduces magic numbers
* Code cleanup

Tested:
* With and without filters
* Disabling and re-enabling filters
* Report a problem > with a bus
* Successfully cancels load more request when:
    * Closing app
    * Locking app 
    * Switching to a different view
* Navigating to view from:
    * Map
    * Bookmarks
    * Trip schedule

### DIFF
--- a/Libraries/aaronbrethorst-onebusaway-iphone-common-fa9a21b/Classes/View/OBAProgressIndicatorView.h
+++ b/Libraries/aaronbrethorst-onebusaway-iphone-common-fa9a21b/Classes/View/OBAProgressIndicatorView.h
@@ -20,7 +20,15 @@
     UILabel * _progressLabel;
     UIActivityIndicatorView * _activityIndicator;
     UIProgressView * _progressView;
+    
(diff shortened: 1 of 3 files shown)


caitbonnar:
Closing this PR because of above discussion and the open issue on the app-modules.

PR #259, opened 2014-02-08
Fixed GIT_COMMIT_SHA for TestFlight build uploads.

neophit:
Modified setBuildVersion.sh to work with the AdHocDistribution configuration which is used to Archive and upload builds to TestFlight.

PR #310, opened 2014-05-23
Improve Google Analytics

### LINKED ISSUE #376: Google Analytics - Changes in region are captured in applicationWillResignActive
While working on [OBA Android Google Analytics implementation](https://github.com/OneBusAway/onebusaway-android/issues/105), we realize that there is two different events for tracking region changes in the app. One where the user switches the region, and another when the region is automatically selected. 

The following code used for tracking the region switches within an action (```configured_region```):
```
- (void)applicationWillResignActive:(UIApplication *)application {
if([self.modelDao.readCustomApiUrl isEqualToString:@""]) {
[OBAAnalytics reportEventWithCategory:@"app_settings" action:@"configured_region" label:[NSString stringWithFormat:@"API Region: %@",self.modelDao.region.regionName] value:nil];
}else{
[OBAAnalytics reportEventWithCategory:@"app_settings" action:@"configured_region" label:@"API Region: Custom URL" value:nil];
}
```
However, in the IOS application lifecyle `applicationWillResignActive` will also be called when the user presses the home button or when the phone receives a call. So, this results lots of same events in the Google analytics console: 

untitled 2

As a solution, ```configured_region``` action could be called where the user switches the region.

bbodenmiller:
Update to latest Google Analytics SDK.

Improve region selection analytics:
* Old implementation recorded set event many more times than it actually
occurred
* New implementation records only when user selects region
* If region changes automatically on startup that is recorded in a
separate already existing event
* Fixes #376 

Move app settings analytics to appDidBecomeActive:
* Improves accuracy of app settings analytics as before Google Analytics would store events to be sent during the next session but if the next session did not occur on the same day Google Analytics would throw out the events.


caitbonnar:
@bbodenmiller Still interested in this PR?

bbodenmiller:
Yes but I'm without a development environment right now. Hope to get that addressed in the next week.

aaronbrethorst:
@bbodenmiller any updates?

bbodenmiller:
@aaronbrethorst currently stuck on https://github.com/OneBusAway/onebusaway-iphone/pull/259#issuecomment-74400162. Would like to get that fixed but if no one has any ideas I'll try to get around it so I can work on this PR.

barbeau:
It looks like this PR would fix https://github.com/OneBusAway/onebusaway-iphone/issues/376 that we noticed on iOS when testing OBA Android analytics.

bbodenmiller:
@aaronbrethorst ready for review. This improves iOS analytics but does not address the comments in https://github.com/OneBusAway/onebusaway/issues/5 which are still ongoing and I will address in a seperate PR.

aaronbrethorst:
Looks good!

Issue #340, opened 2014-11-19
"Out of Region" message always popping up 


caitbonnar:
Received this report from @bradyt:

> On each opening of OneBusAway, it suggests I am not in the Puget Sound area and asks if I'd like to go there.
> Even if I am in Downtown Seattle. Should it wait a moment for the map to initialize?

@Wizecoder, do you think this might have to do with the workaround you came up with? Could you track this issue?

@bradyt, could you quickly verify that your location services are on for OneBusAway? Quick way to do this is to just see if the current location dot is showing on the map; or you can go into the phone settings (Settings>Privacy>Location Services>OneBusAway) and make sure it isn't set to off. Thanks!

PR #344, opened 2014-12-17
Merge 2.3.1 into master

caitbonnar:
I'm releasing a hotfix for the out of region alert always popping up for some users (#340). I've tested the fix with 3 of the users who originally reported the issue and all say it is fixed now, so I believe this is ready for immediate release. (Plus we won't be able to submit any apps during Dec 22-29 due to some maintenance on Apple's end.)

Will pull this in once it is submitted.

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.3</string>
+	<string>2.3.1</string>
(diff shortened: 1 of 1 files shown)


bbodenmiller:
Good news: Since this was released (Jan 2), Out of Region alerts are down significantly:

image
image
image

barbeau:
Nice!

caitbonnar:
I noticed that as well. Thanks, @Wizecoder! :)

Wizecoder:
Phew, glad it worked!

PR #348, opened 2015-01-23
provide more debugging information in contact us

bbodenmiller:
Add:
* Set Region Automatically
* Region
* Custom API

### DIFF
--- a/view_controllers/OBAContactUsViewController.m
+++ b/view_controllers/OBAContactUsViewController.m
@@ -143,9 +143,11 @@ static NSString *kOBADefaultTwitterURL = @"http://twitter.com/onebusaway";
                         controller.mailComposeDelegate=self;
                 				[controller setToRecipients:[NSArray arrayWithObject:contactEmail]];
                 				[controller setSubject:NSLocalizedString(@"OneBusAway iOS Feedback", @"feedback mail subject")];
-                				[controller setMessageBody:[NSString stringWithFormat:@"<br><br>---------------<br>App Version: %@<br>Device: \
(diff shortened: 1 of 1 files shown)


caitbonnar:
These additions look handy. Have you tested and found that it works as expected?

caitbonnar:
Also, I think we should have a pop-up or some sort of notice to the user that this information is going to be added at the bottom of the email message, and inform them that they can delete it if it isn't relevant or they don't want it shared. It would be good to make sure we have their consent since we are including their current location, which users could feel is sensitive information, even if it is intended to just go to us for debugging purposes. (I probably should have thought of this earlier since it's been in there for awhile.)

barbeau:
Right now we have the following info that's auto-populated in the "Contact Us" email body on Android:
* App version
* device model
* Android OS version
* location - provider, lat/long, how many seconds ago it was acquired

It would be good to add an explicit "Please delete the below information if you don't want to share it" message to the email above this info.  I think most users get it, as I've seen a number of feedback emails that have been censored by the user before they send it.

I think adding the additional info that @bbodenmiller proposes is ok.  These emails are going to the admins of the local OBA region, and (despite my earlier comments a while back) I think its good to spell out which region the phone is attached to.

@bbodenmiller or @caitbonnar Could you paste the final format of what you end up using here, just so we know exactly what gets sent?

bbodenmiller:
@caitbonnar I don't have a way to test these on a device right now or in the foreseeable future so if you could that would be great. I added some text about how the user can remove content they do not want to share.

bbodenmiller:
@caitbonnar @aaronbrethorst fixed.

aaronbrethorst:
Looks :+1:

PR #351, opened 2015-01-27
Attempt to fix an issue with the times being off in the schedule list view

icecrystal23:
The times shown on the Trip Schedule view always have the correct minutes, but the hours are off.  This is my attempt to fix it, however I can't seem to reproduce the issue in the simulator so I can't tell if this fix works or not.

### DIFF
--- a/view_controllers/TripDetails/OBATripScheduleListViewController.m
+++ b/view_controllers/TripDetails/OBATripScheduleListViewController.m
@@ -53,7 +53,7 @@ typedef enum {
         _timeFormatter = [[NSDateFormatter alloc] init];
         [_timeFormatter setDateStyle:NSDateFormatterNoStyle];
         [_timeFormatter setTimeStyle:NSDateFormatterShortStyle];
-        [_timeFormatter setTimeZone:[NSTimeZone systemTimeZone]];
+        [_timeFormatter setTimeZone:[NSTimeZone localTimeZone]];
         
         CGRect r = CGRectMake(0, 0, 160, 33);
         _progressView = [[OBAProgressIndicatorView alloc] initWithFrame:r];
--- a/view_controllers/TripDetails/OBATripScheduleListViewController.m
+++ b/view_controllers/TripDetails/OBATripScheduleListViewController.m
@@ -52,7 +52,8 @@ typedef enum {
         
         _timeFormatter = [[NSDateFormatter alloc] init];
         [_timeFormatter setDateStyle:NSDateFormatterNoStyle];
-        [_timeFormatter setTimeStyle:NSDateFormatterShortStyle];        
+        [_timeFormatter setTimeStyle:NSDateFormatterShortStyle];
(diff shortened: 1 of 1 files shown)


caitbonnar:
Hey @icecrystal23, thanks for the contribution! Could you perhaps provide a screenshot about what you mean here? (I don't think I've ever noticed this myself, but it's probably because I'm in Puget Sound.) Once you do I can test it for you. :)

icecrystal23:
Sure, it happens almost 100% of the time for me

icecrystal23:
image
image

caitbonnar:
@icecrystal23 Ah yes - I have seen this before, actually. Thanks! I'll test it locally and pull in if all looks good. :)

barbeau:
Thanks @ <https://github.com/icecrystal23>icecrystal23
<https://github.com/icecrystal23> for the contribution!  Also, just a
friendly reminder to fill out the ICLA if you haven't already.

caitbonnar:
Unfortunately still seeing the error. :( 

img_6932
img_6933

I wonder if it has something to do with the times coming from the API. @aaronbrethorst Do you know what might be happening here?

aaronbrethorst:
I'd think `-localTimeZone` would do it, but I'm guessing there must be another date formatter in use for that. Or something. I'd recommend auditing our use of date formatter objects and seeing where else they pop up in the code base.

aaronbrethorst:
@icecrystal23 - have you had a chance to look into this anymore? No worries if you haven't, but if that's the case, I'd love to get detailed repro steps for this, since I can't find the one stop listed above that demonstrates this issue. Thanks!

icecrystal23:
I didn't look into it more. 

It doesn't need to be a particular stop. It seems to be happening more for agencies which don't have real time data (Sound Transit, Community Transit). I was just looking at stops on 4th Ave in downtown Seattle right now; the first arrival showed the correct times, but any subsequent arrivals have the wrong hour on the stop details list.

aaronbrethorst:
Sorry, I’m still not seeing it. Can you give me a specific stop I can look at?

Also, do you set the time on your iPhone automatically from the network, or do you manually configure it?

icecrystal23:
Automatically. 
Stop #670 in the Puget Sound. Route 512. 
I was having trouble reproducing the issue in the simulator, so you might need to debug on a device.

aaronbrethorst:
Ooh, I bet that’s it (device vs. simulator). thanks!

aaronbrethorst:
@icecrystal23 yep, that was it. Thanks! I'll investigate.

aaronbrethorst:
It appears that for scheduled arrivals—I guess not surprisingly—there's no trip status object on the tripDetails object in `OBATripScheduleListViewController.m` (specifically relevant on line 313 of that file).

When `_tripDetails.status` is `nil`, the date that is loaded into the schedule list view controller is actually in 1970, not 2015. When `status` is represented by an object, the date is correctly in 2015.

icecrystal23:
Cool, thanks for digging into it and finding the root cause!

PR #360, opened 2015-02-23
Show correct Trip Details times for 'scheduled arrival' buses.

### LINKED ISSUE #359: Trip Details times are incorrect for scheduled arrival times
As reported by @icecrystal23 in #351

aaronbrethorst:
Fixes #359

It appears that for scheduled arrivals—I guess not surprisingly—there's no trip status object on the tripDetails object in OBATripScheduleListViewController.m (specifically relevant on line 313 of that file). When _tripDetails.status is nil, the date that is loaded into the schedule list view controller is actually in 1970, not 2015. When status is represented by an object, the date is correctly in 2015.

This commit fixes that issue by properly setting a base date (now) for cases where there isn't a status object. Also adds some debug info to OBATripStopTimeV2.

@icecrystal23 - this fixes the bug you reported.

### DIFF
--- a/Libraries/aaronbrethorst-onebusaway-iphone-common-fa9a21b/Classes/Model/V2/OBATripStopTimeV2.m
+++ b/Libraries/aaronbrethorst-onebusaway-iphone-common-fa9a21b/Classes/Model/V2/OBATripStopTimeV2.m
@@ -8,4 +8,8 @@
     return [refs getStopForId:self.stopId];
 }
 
+- (NSString*)description
+{
+    return [NSString stringWithFormat:@"<%@: %p> :: {arrivalTime: %@, departureTime: %@, stopId: %@, stop: %@}", self.class, self, @(self.arrivalTime), @(self.departureTime), self.stopId, self.stop];
+}
 @end
--- a/view_controllers/TripDetails/OBATripScheduleListViewController.m
+++ b/view_controllers/TripDetails/OBATripScheduleListViewController.m
@@ -50,6 +50,7 @@ typedef enum {
         _timeFormatter = [[NSDateFormatter alloc] init];
         [_timeFormatter setDateStyle:NSDateFormatterNoStyle];
         [_timeFormatter setTimeStyle:NSDateFormatterShortStyle];
+        [_timeFormatter setTimeZone:[NSTimeZone localTimeZone]];
 
         CGRect r = CGRectMake(0, 0, 160, 33);
         _progressView = [[OBAProgressIndicatorView alloc] initWithFrame:r];
@@ -313,8 +314,14 @@ typedef enum {
         serviceDate = status.serviceDate;
         scheduleDeviation = status.scheduleDeviation;
     }
+    else {
+        serviceDate = [[NSDate date] timeIntervalSince1970];
+        scheduleDeviation = 0;
+    }
 
-    return [NSDate dateWithTimeIntervalSince1970:(serviceDate / 1000 + stopTime + scheduleDeviation)];
+    NSTimeInterval interval = serviceDate / 1000 + stopTime + scheduleDeviation;
+
+    return [NSDate dateWithTimeIntervalSince1970:interval];
 }
 
 - (UITableViewCell *)tableView:(UITableView *)tableView loadingCellForRowAtIndexPath:(NSIndexPath *)indexPath {


aaronbrethorst:
@icecrystal23 - can I get you to check out this branch and verify that it fixes the problem you reported?

icecrystal23:
Is there a TestFlight app or similar that I can download to test on a device?

aaronbrethorst:
@caitbonnar what would be the best way to get an ad hoc build our for someone at the moment? Old testflight still?

caitbonnar:
@aaronbrethorst Yep, or emailing them the IPA if they know how to install through iTunes.

caitbonnar:
Oh I'm guessing @icecrystal23 does know how to install through iTunes. :) So whatever's easiest for you! (I just looked at the context of Aaron's question, haha.)

aaronbrethorst:
@caitbonnar I don't think I have the necessary certificate to do ad hoc signing. This will require me to be added to the iTC account for OBA, right?

icecrystal23:
I wouldn't be able to install an ad hoc build unless my device was on the developer portal. I'd need an enterprise or TestFlight build.

aaronbrethorst:
ok. I'm just going to go ahead and merge this, then, unless anyone has strong objections, and—hopefully—we can validate on-device soon. Thanks again!

caitbonnar:
@aaronbrethorst Ah, I see -- yeah, you don't yet have the credentials. :( Since you just merged this in, I will sign the build and upload the current dev branch to TestFlight. @icecrystal23 [Here is the link](http://tflig.ht/1ac8oEg) for you to request to be a beta tester. (I'll add you to the iTunes connect list as well for the future.)

aaronbrethorst:
tyvm :) it'll be good to land all of the changes in the past few days in a new beta build, anyway.

caitbonnar:
@aaronbrethorst Should I release it to all beta testers, then?

aaronbrethorst:
I think that would be terrific. The networking stack changes that hit develop recently are pretty far-reaching, and I’d love to get as much feedback as possible.

caitbonnar:
Ok, sounds good! I'm going to see if I can figure out how to release with the new prelease module in iTC, could you summarize the changes for this beta to include in release notes (if it has em)?

aaronbrethorst:
* Show correct Trip Details times for 'scheduled arrival' buses.
* Searches with ampersands (&) and slashes (/) now work correctly.
* Provide more debugging information in 'Contact Us'.
* Rewrite of much of the core networking stack of the application.

aaronbrethorst:
(and thanks!)

caitbonnar:
Thanks! Will let you know when it's out.

aaronbrethorst:
:+1:

icecrystal23:
Hmm. So I just installed the TestFlight app and it seems that the issue still exists (and potentially even made it worse). I will attach a screenshot demonstrating the issue. The arrival at the stop was at 9:33pm, so the time at the next stop should've been like 9:38 or something.

icecrystal23:
image

aaronbrethorst:
@caitbonnar has a new build been published on TF?

aaronbrethorst:
So it has. I checked the same stop and the data is showing up correctly on my installation. Not sure what might be going on :-/
image

icecrystal23:
The first arrival at the stop tends to works better, remember? (Although that's not always true; I tried another stop somewhere and the first one was all wrong too.)

barbeau:
Assuming this is pulling data from [trip-details](http://developer.onebusaway.org/modules/onebusaway-application-modules/current/api/where/methods/trip-details.html) endpoint - it's possible that this is a server-side error.  We don't use trip-details endpoint on OBA Android, so this wouldn't have been visible in that app.  Furthermore, Sound Transit uses several instances of OBA servers behind a load balancer.  It's very possible that you're actually hitting different servers and getting different responses.  What makes this even more difficult is that the load balancers are configured for "sticky routes", which means that if your public-facing IP is stable, you'll always hit the same server.  So, if you get a bad response the first time, you're likely to always get a bad response from the same server - vice versa for good responses.

I think the best way to eliminate this as a possible error source is for @icecrystal23 to try and capture the server response that's resulting in the bad times, and @aaronbrethorst to do the same (when seeing good times), and compare.

icecrystal23:
I actually was able to reproduce this in the simulator tonight.  I have a new change to try (seemed to work for the stop in the simulator anyway).  I will see if I can get it into a pull request.

bbodenmiller:
It looks like the latest TestFlight build was pushed as 2.4 and I think Google Analytics is reporting it as 2.4. In the future I think it'd be best to make sure test builds have unique versions from release versions so we don't blurr the analytics. I believe at one point we were using the Git SHA1 as part of the version number. It might even be worth looking in to using a seperate tracker for test builds.

PR #362, opened 2015-02-26
Remove all of the old, unused NSManagedObject-based models from the project

aaronbrethorst:
We don't actually use Core Data any longer. No reason to keep these unused files in the project.


bbodenmiller:
:+1 for excess code cleanup

aaronbrethorst:
thanks, @bbodenmiller !

PR #364, opened 2015-02-26
Fixes #355: bug where bookmarks in groups don't allow you to change their names

### LINKED ISSUE #355: Bookmark name changes don't 'stick' if the bookmark is in a group
Reported by a user:

> Hello! I wanted to inform you that if you have bookmarks in a group, and then rename the bookmark, the name will not stick permanently - it will be changed until the app is restarted or reloaded, and then will revert to its pre-edit name.

aaronbrethorst:
- Also, remove a bunch of unnecessary methods where we were statically declaring table views to have 1 section apiece. This is implicit behavior for UITableViews.

Fixes #355


aaronbrethorst:
@bbodenmiller - love a CR if you have a sec

bbodenmiller:
Looks good but I can't say that I've actually tested it.

PR #365, opened 2015-02-26
Fixed a linker error while building a unit test target

nullp0inter:
Added OBAURLHelpers.m file to the test target. Used to fail with linker error like: "Undefined symbols for architecture x86_64: _OBJC_CLASS_$_OBAURLHelpers"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,9 @@ before_install:
 before_script:
     - chmod -R +x ./scripts
     - echo -e "Host github.com\n\tStrictHostKeyChecking no\n" >> ~/.ssh/config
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
We shouldn't need to add any m files to the test target. Which version of Xcode are you using?

nullp0inter:
Xcode Version 6.1.1

nullp0inter:
Reproducible for AppStoreRelease scheme only. Travis uses the Debug one. That is why the CI is green.

bbodenmiller:
How can we update Travis to pickup this error?

nullp0inter:
Changed the Travis CI script accordingly.

aaronbrethorst:
Merged in #371

PR #370, opened 2015-03-02
remove dSYM upload to TestFlight

bbodenmiller:
Remove obsolete release process. I'm unsure if the new TestFlight might require something similar but this process definitely no longer works.


aaronbrethorst:
Thx!

PR #378, opened 2015-03-03
Fix develop to be able to merge to master

bbodenmiller:
Fixed git merging issues mentioned at https://github.com/OneBusAway/onebusaway-iphone/pull/377#issuecomment-76869208.

Verify fix at: https://github.com/OneBusAway/onebusaway-iphone/compare/OneBusAway:master...bbodenmiller:fix-develop?expand=1
Verify that it is currently broken at: https://github.com/OneBusAway/onebusaway-iphone/compare/OneBusAway:master...OneBusAway:develop?expand=1

PR #379, opened 2015-03-03
Remove StopInfo survey

bbodenmiller:
Dependent on/merges #378.

Just preparing for when the survey is no longer needed.

Todo:
* [X] Remove StopInfo Survey
* [ ] Rebase


caitbonnar:
You can remove it now whenever a new version goes out -- we've received over 650 responses due to the incorporation into the app, so we should be good now. Thanks! :)

PR #387, opened 2015-03-16
Increase the number and verbosity of compiler warnings

aaronbrethorst:
The purpose of this pull request is to enable several new compiler warnings, and then fix all of the new warnings that were generated by the compiler. I also fixed a performance issue where we continually created new `NSDateFormatter` objects, which is a very expensive operation.

Getting our codebase in tighter shape in advance of the Hack the Commute hackathon this coming weekend should help ensure that any prospective code changes meet a higher quality bar than they might otherwise.

Lastly, the weakify/strongify macros make it easier for us to properly implement weak and strong copies of `self` for use in blocks.

- Enable several additional compiler warnings
- Fix a number of issues detected through the new, more aggressive warnings
- Add weakify/strongify macros from libextobjc for easier memory management of self in blocks
- Remove several deprecated method implementations
- Add a new date helpers object to vend a specific style of NSDateFormatter that we use regularly


bbodenmiller:
In general looks good but I didn't test it.

Did you by chance run `Uncrustify` on all the files (probably better as a separate PR)?

Additionally I think it would be good to update Travis such that it will return a failure if new warnings are generated rather than only failing for errors.

aaronbrethorst:
I haven't run uncrustify on it. It's having some issues with Xcode 6.2, but I'll get that worked out. Also, we should turn on "treat warnings as errors." That would obviate the need to separately treat warnings as errors in Travis.

PR #391, opened 2015-03-20
remove unused OneBusAway-iPhone IRC channel

bbodenmiller:
Does not appear to be used & Gitter looks to be much better.


aaronbrethorst:
:+1:

bbodenmiller:
@aaronbrethorst do you think you'll have a chance to enable Gitter [GitHub, Travis CI, & Huboard integrations](https://gitter.zendesk.com/hc/en-us/articles/200176722-Integrations) for the hackathon?

barbeau:
Let me know if you get the Travis hooks working for Gitter.  A few builds showed up in OBA Android Gitter room but then went silent.  Not sure if I have it configured right.

aaronbrethorst:
@barbeau check the value of the `on_success` key. I bet you that if it's set to `change`, it'll go quiet until there's a change in the success status. I'm setting iOS' to `always`

barbeau:
IIRC it's ´change´, but build failures is ´always´, and those weren't
showing up either.  And neither were successes after failures.  But that's
a good idea to troubleshoot at any rate, will give it a shot.

aaronbrethorst:
Hmm, bummer. No idea, then :(

PR #392, opened 2015-03-20
Gitter/travis integration

aaronbrethorst:
like it says on the tin, as it were.


barbeau:
@aaronbrethorst you don't have to configure anything else in the Travis console for this, right?  This addition to the script should be enough (in addition to flipping the "Travis" switch in Gitter)?

aaronbrethorst:
That's my understanding...

barbeau:
Hmmm...I wonder if maybe I accidentally flipped the Travis switch more than
once in Gitter and the webhook URL changed...will have to investigate
further on Android...

barbeau:
For the record, it looks like clicking on the "Travis integration" button in Gitter more than once caused the problem on OBA Android.  The second click generated a new webhook URL, and even though both Travis integration records were shown in Gitter UI, the first webhook URL no longer worked.  I deleted both existing Travis integration records and created a new one, and that webhook URL seems to be working now - Travis build updates are again showing up in OBA Android Gitter chat room.

Issue #352, opened 2015-01-30
Add another tracker for Google Analytics unique to OBA iOS


barbeau:
While working on [OBA Android Google Analytics](https://github.com/OneBusAway/onebusaway-android/issues/105#issuecomment-71897488), we realized that Google's [Best Practices](https://support.google.com/analytics/answer/2587087) for tracking the same app on multiple platforms has changed.  Originally, they suggested using the same tracking ID on the various platforms.  Now, they are recommending using different tracking IDs per platform.  However, the catch is that to compare two apps using different tracking IDs (i.e., "rolled up reporting"), you need a subscription to Google Analytics Premium, which costs money.

To make OBA GA future proof, we should add another tracking ID to OBA iOS that tracks only iOS devices.  Here's our general strategy:

| IOS | Android |
| --- | :-: |
| tracker1 (existing property - actually "OBA SmartPhone" with iOS view) | tracker1 (existing property - actually "OBA SmartPhone" with Android view) |
| tracker2  (for rolled up reporting) | tracker3 (for rolled up reporting) |

So, OBA iOS would continue reporting analytics via the current tracker1 ID, but we would add tracker2 ID as well, via multiple trackers feature:
https://developers.google.com/analytics/devguides/collection/ios/v3/advanced#multiple-trackers

I grabbed a new tracker ID for tracker2:
- tracker 2 - iOS (unrolled) - **UA-2423527-18**

aaronbrethorst:
I am not inclined to do this under a time crunch; instead I'd like to spend some time rethinking how analytics works across OBA for iOS. I'm going to push this into the next release when we're not under a time crunch.

barbeau:
Agreed - we need to figure this out collectively as a project too (or at least between iOS and Android).  We're way over the free tier usage limit in GA, and as I understand it we're losing analytics data because of this.  Assuming no one can come up with the $150k/year fee for GA Enterprise, it looks like we're going to need to switch to GA sampling to get under the free tier threshold.

PR #403, opened 2015-03-26
Analytics Simplification/Correctness

aaronbrethorst:
Add a dedicated analytics method for logging the appearance of a view controller, and audit all of our usage of view controller logging to ensure we do it consistently.

### DIFF
--- a/controls/view_controllers/OBAListSelectionViewController.m
+++ b/controls/view_controllers/OBAListSelectionViewController.m
@@ -30,8 +30,8 @@
 
 - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
-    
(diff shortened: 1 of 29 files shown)


bbodenmiller:
Worth noting there is an automatic tracker option: https://developers.google.com/analytics/devguides/collection/ios/v3/screens#automatic

However it appears it may be [complicated by multiple trackers](https://developers.google.com/analytics/devguides/collection/ios/v3/advanced#multiple-trackers) which we need to implement per https://github.com/OneBusAway/onebusaway-iphone/issues/352.

aaronbrethorst:
Yeah, plus I'm not super-excited about the idea of having to use a subclass of UIViewController `¯\_(ツ)_/¯`

PR #404, opened 2015-03-26
Tighten up analytics, and make it easier to use.

aaronbrethorst:
* Replace raw strings in analytics categories with named string constants.


bbodenmiller:
:+1: 

Would like your comments at https://github.com/OneBusAway/onebusaway/issues/5#issuecomment-77833275.

Would also like your help implementing `customDimensionForIndex` for `Show Experimental Regions	`. I can't seem to find the locations where the experimental setting is read/set. Code for the other custom dimensions is at https://github.com/OneBusAway/onebusaway-iphone/blob/develop/models/dao/OBAModelDAOUserPreferencesImpl.m#L211-L247.

aaronbrethorst:
@bbodenmiller Thanks, Ben. I'll take a look.

PR #405, opened 2015-03-26
Automatically opt in all OBA view controllers to analytics reporting

aaronbrethorst:
I'm WFH and—while eating lunch—was thinking about how we could use method swizzling to simplify our analytics approach. Turns out to be really straightforward, and took me less time to implement than it did to eat :)

- Replace -viewWillAppear: implementations of analytics screen tracking with a method swizzling approach.
- Add VoiceOver tracking to the -reportViewController: analytics method so that we can accurately see all VoiceOver usage of the app.

### DIFF
--- a/utilities/OBAAnalytics.m
+++ b/utilities/OBAAnalytics.m
@@ -19,7 +19,16 @@ NSString * const OBAAnalyticsCategoryUIAction = @"ui_action";
 NSString * const OBAAnalyticsCategoryAccessibility = @"accessibility";
 NSString * const OBAAnalyticsCategorySubmit = @"submit";
 
+NSInteger const OBAnalyticsDimensionVoiceOver = 4;
+
 @implementation OBAAnalytics
+
++ (void)configureVoiceOverStatus {
+    if (UIAccessibilityIsVoiceOverRunning()) {
+        NSString *dimensionValue = @"VoiceOver: ON";
+        [[GAI sharedInstance].defaultTracker set:[GAIFields customDimensionForIndex:OBAnalyticsDimensionVoiceOver] value:dimensionValue];
+    }
+}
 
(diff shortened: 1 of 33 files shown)


utilities/OBAAnalytics.m
```diff
@@ -32,6 +32,10 @@ + (void)reportScreenView:(NSString *)label {
 
 + (void)reportViewController:(UIViewController*)viewController {
     [self reportScreenView:[NSString stringWithFormat:@"View: %@", viewController.class]];
+    
+    if (UIAccessibilityIsVoiceOverRunning()) {
+        [OBAAnalytics reportEventWithCategory:OBAAnalyticsCategoryAccessibility action:@"voiceover_on" label:[NSString stringWithFormat:@"Loaded view: %@ using VoiceOver", viewController.class] value:nil];
```
bbodenmiller: Another option may be to just use a custom dimension that keeps track of if voice over is running or not. This would also allow one to see which features are used with voice over vs. which are not.

bbodenmiller:
Looks good if it works properly. :)

aaronbrethorst:
@bbodenmiller thanks for the suggestion. I've set up VoiceOver as a custom dimension, and automatically set up reporting for it if enabled.

bbodenmiller:
If the user changes their VoiceOver setting mid-session would the app have to be force quit and started again before it would pickup the change? Additionally do you have any issue with me changing the scope on GA to hit rather than user? I ask because I think it the current implementation it [would always show VoiceOver on after they had enabled it](https://developers.google.com/analytics/devguides/platform/customdimsmets#scope), even if they disabled it.

aaronbrethorst:
I expect anyone who changes their voiceover settings mid session is unlikely to be visually impaired and I'm not too concerned about losing that data point. 

Feel free to change the scope. Thanks!

PR #406, opened 2015-03-28
Fixing 404 link in Roadmap.md

kevinvanderlugt:
I was browsing around the project learning how to get started and found this 404 link.  
Easy way to start, would love to get more involved especially if there is a way to bring one bus away to apple watch.

Cheers!


aaronbrethorst:
thanks, @kevinvanderlugt!

aaronbrethorst:
and i'd love to see OBA on the Apple Watch sooner than later, too :)

bbodenmiller:
Would OBA for watch would require similar library compartmentalization as the today widget?

aaronbrethorst:
Yeah, I expect that they'll actually end up using the same extracted code.

PR #410, opened 2015-04-06
improve location prompt

bbodenmiller:
- [x] ask for location access earlier to fix errors
- [x] simplify code since it is now expected that the app will be built
using Xcode 6

- [x] update iOS 8 check for `hasRequestedInUseAuthorization`
- [x] fix bug introduced on iOS <8 where location following does not work


aaronbrethorst:
Looks :+1: so far

bbodenmiller:
@aaronbrethorst ready for review.

aaronbrethorst:
:+1:

PR #414, opened 2015-04-10
Fix Issue 412

frankus:
This is a fix for issue 412 that uses roughly the same logic as the list view "connecting routes" feature for the annotation views.


aaronbrethorst:
neat, thanks! :+1:

PR #416, opened 2015-04-17
Implemented basic views for an Apple Watch app using the Recent and ...

syoung-smallwisdom:
...Bookmarked stops to create a glance and simple app for checking arrival times.

04/25/2015
UPDATE: Revised the code so that it will now run both in simulator and on my statistically significant sample size of one Apple Watch. Note: there were a lot of warnings and build errors that showed up when running against Xcode 6.3.1 that did not appear when running in the simulator on Xcode 6.2

04/24/2015 
UPDATE: The Apple Watch app is not working on device. `-openParentApplication:reply:` is never fires the reply (either with an error or not). I'm looking into it but it may take some time. Leaving the pull request open for now but will likely close this and open it back up when/if I am able to get things working.

---- Original Comment ----

Greetings. I didn't see anything on the list for supporting the Apple Watch, but this is a feature that I want (as do some of my friends and co-workers).

This is a fairly simple implementation that is limited to ONLY displaying information about your bookmarked stops (with filtered routes) and your most recent stop.  The "Glance" looks for the two next buses that will arrive to any stop that is within the bookmark group or recent stop that is closest to you (if current location is allowed) or your most recent stop if it is not.

I made very few changes to the iPhone project, though there were a few that were required in order to support the watch implementation. B/c the project supports iOS 6, I couldn't use shared code without a lot of changes to the iPhone app. To limit those changes I created some simple model objects that I grouped in the Apple Watch project and added the suffix "WK" to differentiate them from the base iPhone app implementations. They use a simple KVO pattern to parse the json.

Anyhow, hope you like it.

ios simulator screen shot - apple watch apr 17 2015 12 48 03 am
ios simulator screen shot - apple watch apr 17 2015 12 48 19 am
ios simulator screen shot - apple watch apr 17 2015 12 48 53 am


Info.plist
```diff
@@ -9,7 +9,7 @@
 	<key>CFBundleExecutable</key>
 	<string>${EXECUTABLE_NAME}</string>
 	<key>CFBundleIdentifier</key>
-	<string>BUNDLE_IDENTIFIER</string>
+	<string>BUNDLE-IDENTIFIER</string>
```
syoung-smallwisdom: Note: this change was necessary to support the Apple Watch project.

OneBusAway_WatchKit_Extension/OBAArrivalsAndDeparturesInterfaceController.m
```diff
@@ -0,0 +1,133 @@
...
+
+    [self.dataTask cancel];
+    self.dataTask = nil;
+}
+
+- (NSArray *)filteredRouteIds {
+    return nil;
+}
+
+- (void)refreshArrivalsAndDeparturesForStop:(OBAStopWK *)stop {
```
beliyd: Is there any reason why this logic would be inside the controller and not the OBA "service layer"?

I'm somewhat sure that we probably already have this call in the main OBA service stack, no? Could be wrong.

aaronbrethorst:
@syoung-smallwisdom - sorry I didn't reply sooner. The PR email seems to have been buried somewhere in my inbox. It looks like the Travis failure will be resolved "soon'ish": https://github.com/travis-ci/travis-ci/issues/3373

I should (fingers crossed) have an   Watch on Friday, and I can—assuming on-device testing is there on day 1—take a better look into this then. Otherwise, 2,000+ LOC is a lot to review. Thanks, by the way, this is awesome!

OneBusAway_WatchKit_Extension/data/OBAStopGroupWK.h
```diff
@@ -0,0 +1,28 @@
...
+
+typedef NS_ENUM (NSInteger, OBAStopGroupType) {
+    OBAStopGroupTypeRecent,
+    OBAStopGroupTypeBookmark,
+    OBAStopGroupTypeNearby
+};
+
+@interface OBAStopGroupWK : OBAModelObjectWK
+
+@property (nonatomic) NSString *name;
```
beliyd: nitpick:  Wonder why not make the ownership explicitly defined.  There used to be a time when LLVM defaulted to assign.

In regards to NSString setters, it is good practices to make them copy, in case someone passes in an NSMutableString

OneBusAway_WatchKit_Extension/OBAArrivalsAndDeparturesInterfaceController.m
syoung-smallwisdom: @beliyd 
I started out trying to pull it into a library and stopped primarily for two reasons:

1. Currently, the main project is targeted at iOS 6.0. Because of that, I couldn't use a dynamic embedded library.

2. It's already a *huge* change just to add what I considered the bare minimum for an Apple Watch app (that wasn't pointless).  I wanted to keep this iteration separated as much as possible. Towards that end, I tried to keep to the existing model (but simplified) so that it would be fairly straight-forward to change to using a shared library, but used a naming convention that ended with "WK".

In the long run, I would agree that a shared services/model library would be a lot cleaner and easier to maintain, but getting there is more than I had the time to tackle. :)

OneBusAway_WatchKit_Extension/data/OBAStopGroupWK.h
syoung-smallwisdom: Good point, been a while since I've created my own model objects for a public use project. I will change that.

syoung-smallwisdom:
@aaronbrethorst 
Thanks for your patience.  Yeah, it's a lot of code.  Worth keeping in mind that a bunch of it is setup files and the "scaled down object model" I'm using as a work-around to not being able to create a shared embedded library targeting iOS 6.  

And yes, really curious to see if this actually works when it's not the simulator.  ;)

aaronbrethorst:
I'd recommend looking into the possibility of building a framework by adding model files both to the framework and OBA targets. It'll result in some binary bloat, but I'd guess that would be a megabyte or three or so at most, which is a totally reasonable outcome as far as I'm concerned.

OneBusAway_WatchKit_Extension/OBAArrivalsAndDeparturesInterfaceController.m
beliyd: I created a separate issue to track modularization of the app.  I agree it makes sense to split the work into two efforts.

Until we go 8.0+ we'll have to settle for shared static library instead of embedded framework :(

bbodenmiller:
I emailed Travis CI and got us access to the Xcode 8.2 test boxes. Simply add osx_image: beta-xcode6.2 to .travis.yml to access the new infrastructure. We're also on a list to be notified when Xcode 6.3 is ready for testing.

syoung-smallwisdom:
Updated .travis.yml file and changed NSString properties to copy.

aaronbrethorst:
@syoung-smallwisdom - when trying to build I'm getting this error:

    error: The value of WKCompanionAppBundleIdentifier in your WatchKit app's Info.plist "BUNDLE_IDENTIFIER" does not match the bundle identifier of your companion app "BUNDLE-IDENTIFIER".


Can you say why you changed that value?

aaronbrethorst:
@syoung-smallwisdom looking good! there are some rough edges here and there to polish, but it's working! https://twitter.com/aaronbrethorst/status/591832686538272768

syoung-smallwisdom:
@aaronbrethorst 
I have re-submitted since it was *not* working on my Apple Watch. I think that the requests were taking too long to fetch.  When I'd originally done this, I'd thought that the requests for stop info were cached but turns out a slow network connection makes a difference.  If the requests take too long, then the `openParentApplication` doesn't reply.

Additionally, I'm tired and adding the script for replacing the BUNDLE_IDENTIFIER and GIT_COMMIT_SHA flags was too much for my brain. I included the GIT_COMMIT_SHA change in a separate commit.

This adventure definitely got more complicated than I'd originally thought it would. Can definitely see a future where a shared library would not only be "nice" but necessary b/c the next "want" would be to add "search for buses near my current location".

beliyd:
@aaronbrethorst @syoung-smallwisdom There is a OSX network link conditioner that you can use to "simulate" adverse network conditions.   I have not tried it with Apple Watch sim, but it may be something of use for us to test with.

syoung-smallwisdom:
@beliyd 
The Apple Watch app actually runs as an extension on the iPhone, so yes, the network simulator will work for the watch as well.

@aaronbrethorst 
So... this was an interesting experiment and I will leave it up for a few days so you can test or pull the code if you would like. I cannot recommend it (as is) for submission for a couple of reasons. First, the number of work-arounds to a shared SDK that I had to do are fairly substantial at this point and should be refactored to be in the shared library. Second, for the life of me, I cannot figure out the magic for getting the Glance to update consistently. It will work for a while and then will stop updating and will show stale data. No idea why and I just don't have the time to look into it right now. Sorry.

aaronbrethorst:
No apologies needed, this is awesome! :) I'll happily help shepherd this through to the App Store. Thanks again for your help. Since you know about ten times more about building WK apps than I do at this point, I may ping you for help.

syoung-smallwisdom:
@aaronbrethorst , @beliyd, Hi guys, thanks for all your help and reviews.

I'm going to close this pull request. I've been testing it for a couple of weeks now and I was able to fix some issues but...

The problem is in the limitations of the Apply Watch. Because the only code that third-party devs have access to is on the iPhone (and not running on the watch itself) there is no way of guaranteeing that the information displayed on the watch's Glance isn't stale. Specifically, when your glance is shown, the watch is *suppose* to send `-willActivate` to the phone, and eventually it does. The phone can react to that and send a message to the phone to clear stale data, and again, eventually it does.

The problem is that word "eventually". Meanwhile, the watch is showing stale data which can be very confusing. The only thing the dev can do is hint at the state, by using a `WKInterfaceDate` element to display the current time. Since this is also stale for a stale view, it will be incorrect and the user could say to themselves, "Wait, it's not 4pm, it's 8am. That's not right". Of course, then the user blames the app for confusing them. *sigh* (I've been reading user reviews that say basically this.)

With v1 of the Apple Watch, I am not really sure how to resolve this so that the UI/UX isn't confusing and inaccurate. Ideally, Apple will provide an option to expire the content and allow the display to change to a loading state prior to displaying that stale data to the user, but currently they do not.

Bit of a bummer. Since you have access to the code, you might play with it a bit yourselves and see what you think.

PR #419, opened 2015-05-01
Changing title causes UI error

cashgl:
If a network error occurs while looking at the map, once you regain
connection and stops load, once you go to look at an individual stop,
the back navigation button title is "Error connecting" instead of "Map"


aaronbrethorst:
Thanks for the PR! Have you filled out the ICLA yet?

cashgl:
Yes I did right before initiating the PR

PR #421, opened 2015-05-10
switch libraries to CocoaPods & upgrade libraries

### LINKED ISSUE #402: Switch to CocoaPods for libraries
Seems like it would be easier to manage additionally frameworks/libraries using CocoaPods. So far I've found the following already in use in the app that have CocoaPods:

* [TWSReleaseNotesView](http://cocoapods.org/pods/TWSReleaseNotesView)
* [libextobjc](http://cocoapods.org/pods/libextobjc)
* [MarqueeLabel](http://cocoapods.org/pods/MarqueeLabel)
* [GoogleAnalytics-iOS-SDK](http://cocoapods.org/pods/GoogleAnalytics-iOS-SDK)

bbodenmiller:
* fixes #402
* also drops unneeded framework references - see
https://stackoverflow.com/questions/24902787/dont-we-need-to-link-framework-to-xcode-project-anymore
* upgrades libraries to latest releases


aaronbrethorst:
Please set explicit versions for the Pods listed in the Podfile; otherwise looks :+1:

aaronbrethorst:
Actually, scratch that. I'd rather get this merged, and I can take care of the change afterwards.

bbodenmiller:
I debated exact versions - seemed overkill given that `Podfile.lock` keeps the exact version and you have to run `pod upgrade` (not just `pod install`) to force an upgrade.

PR #423, opened 2015-05-11
iOS 8-ification of OneBusAway

aaronbrethorst:
- Upgrade the app to version 2.4.0
- Fix deprecations and warnings
- Remove outmoded checks for system version
- Specify explicit versions of Cocoapods
- Change minimum version of iOS in Podfile to 8.0

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.3.2</string>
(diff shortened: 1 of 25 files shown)


bbodenmiller:
Assuming you saw my comment on the last PR & think it's time to drop iOS 7 support then :+1

aaronbrethorst:
I did. Alan and Caitlin are on-board with it, and going iOS 8+ makes it way easier to build Today screen extensions with the built-in support for dynamic frameworks. Thanks :)

PR #424, opened 2015-05-11
Remove survey

aaronbrethorst:
* Depends on #423 
* Obviates #379 

Remove the stop info user survey from the codebase completely. We can always restore it via Git history if/when it's needed again. I don't like keeping unused code lingering in the codebase.


bbodenmiller:
:+1:

aaronbrethorst:
thanks again, @bbodenmiller!

Issue #322, opened 2014-09-22
Add iOS 8 TODAY screen to show a few favorite stop+route intersections


chadsy:
Pull down the today screen, see the next 1-3 buses of a specific route due at a specific stop. Killer feature for commuters who are always on one route at a small set of stops.

aaronbrethorst:
Moving out of 2.4.

beliyd:
This will be unblocked by #417

beliyd:
@Wizecoder @aaronbrethorst I was thinking of messing around with this over next weekend, didn't want to interfere if work is already in progress though.   Thoughts?

aaronbrethorst:
That sounds great, but I'd love to talk about design first.

beliyd:
Something typical, shared user defaults + a simple Storyboard UI that uses same art assets as existing app, perhaps even re-use existing controllers through container controllers.

aaronbrethorst:
Sorry, I should have been more specific: that sound great, but I'd love to talk about the user experience, and what data is exposed to the user in the Today screen.

beliyd:
What would ya all like to be shown?

bbodenmiller:
Personally I'd like the today screen to show the arrival times for the nearby routes that I've favorited. Unfortunately right now however we only track favorites at a stop level & not a route level.

aaronbrethorst:
To second that, I'm in complete agreement with Ben that this is what we should be displaying. So for now, I guess, we can show the next 3(?) buses at the closest stops you've favorited perhaps.

beliyd:
Cool.  I'll see if I can spike it over the weekend, will report back if anything un-expected comes along.

aaronbrethorst:
Thanks!

beliyd:
Started working on this.  

May need to do multiple smaller PRs out first laying the groundwork.

aaronbrethorst:
Awesome :+1:

beliyd:
@aaronbrethorst @bbodenmiller fyi.  The "develop" branch is not building.

error:  ld: library not found for -lPods-MarqueeLabel

I have cocoapods installed.

Any thoughts?

bbodenmiller:
@beliyd assume build issue is fixed as your comment has been deleted? I just did a refresh clone and build from `develop` with no problems.

beliyd:
Got pulled off into more work stuff :(   Unfortunately not looking good for me finishing this :(

[...]

PR #425, opened 2015-05-25
Common code refactoring

beliyd:
Extracting common init logic from OBAApplicationDelegate into a separate class.

Extensions are not allowed to reference the app shared app delegate, so this is a step to get extension support with common init.

Prep for #322


aaronbrethorst:
Generally looks alright. Couple things: please rename `instance` to `sharedApplication` and clarify the comment on the start method. I'll have more feedback later today. Thanks!

beliyd:
Sounds good.  I'm not partial to naming, so will rename to whatever ya all want :)

location/OBARegionHelper.m
```diff
@@ -151,19 +141,16 @@ - (void)setRegion {
 
 
 - (void)locationManager:(OBALocationManager *)manager didUpdateLocation:(CLLocation *)location {
-    OBALocationManager *lm = self.appDelegate.locationManager;
-
-    self.location = lm.currentLocation;
-
-    [self setNearestRegion];
+    self.location = [OBAApplication instance].locationManager.currentLocation;
+   [self setNearestRegion];
```
bbodenmiller: Looks like this needs an extra space in front of it

sdk/sdk/OBAApplication.h
```diff
@@ -0,0 +1,44 @@
+//
+//  sdk.h
+//  sdk
+//
+//  Created by Dima Belov on 4/25/15.
+//  Copyright (c) 2015 One Bus Away. All rights reserved.
```
bbodenmiller: "OneBusAway"

sdk/sdk/OBAApplication.m
```diff
@@ -0,0 +1,120 @@
...
+    
+    OBADataSourceConfig * obaRegionDataSourceConfig = [[OBADataSourceConfig alloc] initWithUrl:regionApiServerName args:obaArgs];
+    OBAJsonDataSource * obaRegionJsonDataSource = [[OBAJsonDataSource alloc] initWithConfig:obaRegionDataSourceConfig];
+    _modelService.obaRegionJsonDataSource = obaRegionJsonDataSource;
+    
+    [userDefaults setObject:appVersion forKey:@"oba_application_version"];
+}
+
+- (NSString *)userIdFromDefaults:(NSUserDefaults*)userDefaults {
+
```
bbodenmiller: Looks like there may be extra lines on a few of these functions.

sdk/sdk/OBAApplication.h
```diff
@@ -0,0 +1,44 @@
...
+
+@property(nonatomic, strong, readonly) OBAReferencesV2 * references;
+@property(nonatomic, strong, readonly) OBAModelDAO * modelDao;
+@property(nonatomic, strong, readonly) OBAModelService * modelService;
+@property(nonatomic, strong, readonly) OBALocationManager * locationManager;
+
+/**
+ *  This block, if set, is called whenever refreshSettings also refreshes a region.
+ */
+@property (nonatomic, copy) dispatch_block_t regionRefreshed;
```
aaronbrethorst: I'd like to see the block removed and replaced with either a delegate method or a notification. Including a block on an asynchronously-executing method call is sensible and easy-to-understand to me, but it quickly becomes thorny when you're setting a single property on a single object, especially since any other object in the system can overwrite the block you originally passed in.

sdk/sdk/OBAApplication.h
```diff
@@ -0,0 +1,44 @@
+//
+//  sdk.h
+//  sdk
+//
+//  Created by Dima Belov on 4/25/15.
+//  Copyright (c) 2015 One Bus Away. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "OBAReferencesV2.h"
```
aaronbrethorst: Replace these with `@class` statements, and move the `#import`s into the .m file.

sdk/sdk/OBAApplication.m
```diff
@@ -0,0 +1,120 @@
...
+@property(nonatomic, strong, readwrite) OBAReferencesV2 * references;
+@property(nonatomic, strong, readwrite) OBAModelDAO * modelDao;
+@property(nonatomic, strong, readwrite) OBAModelService * modelService;
+@property(nonatomic, strong, readwrite) OBALocationManager * locationManager;
+
+@end
+
+@implementation OBAApplication
+
++(instancetype) instance {
```
aaronbrethorst: Rename to `sharedApplication`

sdk/sdk/OBAApplication.m
```diff
@@ -0,0 +1,120 @@
...
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        oba = [[OBAApplication alloc] init];
+    });
+    
+    return oba;
+}
+
+-(void) start {
+    self.references = [[OBAReferencesV2 alloc] init];
```
aaronbrethorst: Why isn't this in the `-init` method?

sdk/sdk/OBAApplication.m
```diff
@@ -0,0 +1,120 @@
...
+    self.modelDao = [[OBAModelDAO alloc] init];
+    self.locationManager = [[OBALocationManager alloc] initWithModelDao:self.modelDao];
+    
+    self.modelService = [[OBAModelService alloc] init];
+    self.modelService.references = self.references;
+    self.modelService.modelDao = self.modelDao;
+    
+    OBAModelFactory * modelFactory = [[OBAModelFactory alloc] initWithReferences: self.references];
```
aaronbrethorst: space after the `: `

aaronbrethorst:
Looks :+1: modulo the feedback above. Let me know when you're ready for another peek. Thanks!

sdk/sdk/OBAApplication.m
beliyd: I'll run the formatter when other things are fixed.

sdk/sdk/OBAApplication.m
beliyd: In case there is a need to async init.  I'll update the doc mentioning that only "instance" / "sharedApplication" should be used.  Same reason one would not try to re-init app delegate.

sdk/sdk/OBAApplication.h
beliyd: These imports are needed to here for the same reason they were in OBAApllicationDelegate.  Sure, we could do that, but then we'd have to add a bunch of imports everywhere else.  Was out of scope for this change.

sdk/sdk/OBAApplication.h
beliyd: I disagree.  This blocks only purpose is to maintain legacy business logic compatibility.   I'd like to see it gone altogether but did not feel a refactor was justified for this PR.  The reason it is still here is because of how coupled OBARegionHelper is with non-extension compatible APIs and could not be extracted easily into another layer.

Sure adding delegation / event system support will make it easier to expand in future, but that is really not the point here.   A delegate can be overwritten just as easily (I can add this block as a param to the factory method if you wish)

Additionally it will add another layer of indirection to the codebase, that is simply not needed to achieve this.

imho, there is no point in trying to be fancy here, this block achieves the requirements with minimal impact.

As far as sync or async, the consumer of this block is abstracted from that detail.  For all we know this class is all async.

sdk/sdk/OBAApplication.h
beliyd: darn xcode... can't rename a file properly :)  I'll fix.

sdk/sdk/OBAApplication.h
aaronbrethorst: > This blocks only purpose is to maintain legacy business logic compatibility

Here's my perspective on this: this code is still going to be here in three years. Some day, someone will come along and add code that overwrites that copied block. Maybe it'll slip through a code review, and then the app will unexpectedly stop working correctly in a way that will be extremely difficult to track down.

> that is really not the point here

The point for me is to ensure that whoever works on OneBusAway after us at some point in the future doesn't unexpectedly shoot themselves in the foot. Let's do the right thing not just for today but also for tomorrow.

> A delegate can be overwritten just as easily

True, so let's use notifications. Or, if you're dead-set on using blocks, change the `regionRefreshed` property to be a pair of methods called something like `-addRegionRefreshedHandler:` which will vend an opaque token of some kind that is later passed to `-removeRegionRefreshedHandler:` to remove that block.

beliyd: Sure.  I'll replace it with NSNotification, seems like can be done with about same LOC using the block type notifications.

aaronbrethorst: thx, looking forward to integrating this! :+1:

beliyd:
Feedback addressed.

OBAApplicationDelegate.m
```diff
@@ -52,103 +52,48 @@ - (id)init {
...
         _stopIconFactory = [[OBAStopIconFactory alloc] init];
-        
-        [self refreshSettings];
+
+
+        self.regionObserver = [[NSNotificationCenter defaultCenter] addObserverForName:kOBAApplicationSettingsRegionRefreshNotification
+                                                                                object:nil
+                                                                                 queue:[NSOperationQueue mainQueue]
+                                                                            usingBlock:^(NSNotification *note) {
+                                                                                self.regionHelper = [[OBARegionHelper alloc] init];
```
aaronbrethorst: retain cycle

beliyd: Correct.   

The block captures self, the app delegate, which is really a singleton that lives for the lifetime of the application.    If you feel strongly about this I'll weakify it, but it doesn't make a difference one way or another.

sdk/sdk/OBAApplication.h
```diff
@@ -0,0 +1,45 @@
+//
+//  OBAApplication.h
+//  OneBusAwaySDK
+//
+//  Created by Dima Belov on 4/25/15.
+//  Copyright (c) 2015 One Bus Away. All rights reserved.
```
barbeau: Legal nit - this should be `Copyright (c) 2015 Dima Belov`, and should include the Apache 2.0 license header.  You own the copyright because you created the code, and then you license your work via Apache, and to the OneBusAway project via the ICLA (via UW for now, since we don't yet have a legal "OneBusAway" entity).  Same applies to any other similar files.

beliyd: I didn't actually write like 99% of the code, just extracted it, still applies?

barbeau: If it includes others code technically it should include all authors and
the year range:
´Copyright (c) 2011-2015 Brian Ferris, x, y, ..., Dima Belov´

I'd do the best you can to assemble this given the Git history.

ui/trip_details/OBATripScheduleMapViewController.m
```diff
@@ -61,29 +61,37 @@ - (void)viewWillAppear:(BOOL)animated {
...
-                                              [self.progressView setMessage:NSLocalizedString(@"Downloading...", @"message") inProgress:YES progress:progress];
-                                          }
-                                          else {
-                                              [self.progressView setInProgress:YES progress:progress];
-                                          }
-                                      }];
+                                        progressBlock:^(CGFloat progress) {
+                                            @strongify(self);
+
+                                            if (progress > 1.0) {
```
aaronbrethorst: It's out of scope for this pull request, but the uncrustify config needs to be fixed to prevent stupid stuff like this. A dedication to a particular line length at the expense of legibility doesn't help us.

aaronbrethorst:
@beliyd once you've addressed @barbeau's feedback, I'll merge the PR.

beliyd:
Will update tmrw and fix the retain cycle.  Could not get to it today.

aaronbrethorst:
Sounds good, thank you!

beliyd:
btw: git shortlog -sn -- OBAApplicationDelegate.m  Might be useful for future stuff.

aaronbrethorst:
@beliyd have you addressed @barbeau's feedback?

beliyd:
@aaronbrethorst just pushed changes discussed, added copyright to the new file added.   I won't have time to add @copyright to all modified files tonight though... I can do that tomorrow if need be.. sorry for the delay, been a crazy couple days here.

aaronbrethorst:
@beliyd all good. I'll merge as soon as Travis comes back green, and update the (c)'s when you have time. thanks again!

beliyd:
@aaronbrethorst thanks.  I'll add a note to myself to do it for next PR in this iteration.   Next weekend :)

PR #427, opened 2015-07-02
Fixes #413 — Blurry pins on iOS 8.3+

### LINKED ISSUE #413: Map pins are blurry on iOS 8.3 (iPhone 6)
Looks like something changed in iOS 8.3, causing us to use 1x assets(?) on the map for our bus stop icons instead of proper retina assets. Yuck.

img_0637

yogipatel:
The stop icon factory was being instantiated *way* too early in the application's lifecycle. Loading images too early caused the UIScreen's scale factor to not be properly set (it was reporting as 1.0 on a retina device). In general we shouldn't really be overriding the application delegate's init method, as this is run before the underlying infrastructure is even spun up (run loop, window, etc.). There must've been a change in the UIScreen initialization sequence in iOS 8.3.

Additionally, the OBAStopIconFactory was unnecessarily caching the UIImage objects for each icon. [UIImage imageNamed:] is already cached by the system.

There's some additional clean-up that can be done here — using enums for the different stop types, using a singleton instead of a factory (that is essentially what is happening anyway), etc. I'll submit another PR for those once this is merged.

### DIFF
--- a/OBAApplicationDelegate.m
+++ b/OBAApplicationDelegate.m
@@ -54,8 +54,6 @@ static NSString *const kAllowTracking = @"allowTracking";
     if (self) {
         self.active = NO;
 
-        _stopIconFactory = [[OBAStopIconFactory alloc] init];
-    
         @weakify(self);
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
awesome, thanks!

Issue #349, opened 2015-01-23
Create a "Donate" link in the mobile app


barbeau:
This would be for donations to the overall OBA project (not regions), and would direct to this link:
http://onebusaway.org/donate/

Currently the donation account is managed via UW, and used for overall project expenses.

Also ticketed for Android in https://github.com/OneBusAway/onebusaway-android/issues/201.

caitbonnar:
Cool! Will add to the Ready list. :)

barbeau:
Thanks @caitbonnar!  I haven't determined exactly how this would be shown in OBA Android (I assume in Help or Preferences lists), so please keep me posted for how it ends up being shown in OBA iOS.

alanborning:
For the iPhone, how about if we put it at the top level under "Info".

caitbonnar:
@alanborning That sounds like a good place to me, too. 

@barbeau Whoever works this in should post a screenshot so you guys can see what it looks like in iOS.

aaronbrethorst:
fixed with #429

PR #429, opened 2015-07-03
Implements #349. Adds a donate button to the info list.

yogipatel:
Additionally, a bit of minor clean up + moved the URL strings to constants. Removed "localization" on the URL strings too, because that makes no sense. Note that the current privacy policy URL is returning a 404.


aaronbrethorst:
awesome

barbeau:
Thanks Yogi!  The main OBA.org site underwent a transition to WordPress,
and it looks like the privacy page got lost in the shuffle.  I'll see if I
can fix that early next week.

PR #430, opened 2015-07-03
Replaced the location and list icons with nicer ones.

yogipatel:
Thought the default ones were a bit blurry and out of place. These icons are extracted from iOS itself, so shouldn't need any attibution (I think). Includes `@3x` images.


yogipatel:
Ideally it would be nice to replace the location button with a `MKUserTrackingBarButtonItem`, but that would change the current behavior. `MKUserTrackingBarButtonItem` allows for map tracking with and without heading (which would probably expose the rotation bug). Currently the button just updates the location.

bbodenmiller:
Do you have a before and after screenshot?

yogipatel:
New:
new_icons
Old:
old_icons

bbodenmiller:
Looks good. Do you happen to be interested in also tackling the fact that the location button always shows as solid even when not tracking the users current location?

yogipatel:
Sure, #432.

PR #437, opened 2015-08-22
Fix issue where table header text overlaps with cell text

### LINKED ISSUE #436: Table header text overlaps with cell text when attempt to scroll
Actual | Expected
---------|------------
img_5715|img_5718

I found this from #43, but only from [the comments](https://github.com/OneBusAway/onebusaway-iphone/issues/43#issuecomment-23411652)

tomtclai:
Fixes #436. 

Before | After
---------|------------
img_5715|img_5718

 I can make it even more opaque, if you prefer.


aaronbrethorst:
Awesome!

Issue #65, opened 2013-07-30
more map options (satellite, 3d, etc.)


bbodenmiller:
we should look in to more map display options such as satellite, flyover, 3d.... whatever else apple maps offers.

ualch9:
Began working on this [here](https://github.com/ualch9/onebusaway-iphone/commits/mapsRevamp)

PR #439, opened 2015-08-22
add satellite / hybrid map option #65

tomtclai:
#65 

img_5721| img_5722
-------- | --------
img_5723| img_5724


bbodenmiller:
Hmm. I'd prefer to more easily be able to change this on the fly from the map view.

tomtclai:
That makes sense. What should it look like? should we use the existing button on the top right, make it take you to a table that includes satellite/hybrid etc and the existing list view as option? I think we'll need a different icon if we go that route.

bbodenmiller:
Good question, @aaronbrethorst any thoughts? Might be good to check the analytic data for usage of that top right button. Can't say I've ever used that button.

Somewhat unrelated: I'd also like to see the location button on the left updated to be like the one in Apple Maps which properly shows if it's currently following your location or not.

aaronbrethorst:
I'm a little leery of removing the list option from the map view for now, as I expect it's regularly used by visually impaired users. I think the entire map view (well, really the entire app Ux) needs a rethink. Location button: totally agree.

I'd be interested in exploring a toggle like the one that Apple used to use to toggle the flash on and off in their camera app. We could stick something like that into the bottom left corner of the map to let users toggle satellite and map views. We'll need a satellite icon, too, and I have an email out to the makers of this icon set: https://icons8.com/web-app/5374/satellite. They seem to allow open source app creators to use their icons for free.

Thoughts?

tomtclai:
Here's a mock up, is this what you're thinking? 

|          |             |
| ------------- |-------------|
|sattelite button|sattelite button tapped|

tomtclai:
As an idea, apple maps use a segmented control for this. But no idea where you would put it on OBA.
img_5737

aaronbrethorst:
@tomtclai Right, there's no place to stick the segmented control...for now at least. We need to do something of a rethink in terms of the overall design of the app.

Also, basically yes. The button should be ~40px on a side, and I think that the orientation of the satellite icon from that set looks weird (it's pointing *out* of the app instead of into it, but otherwise you're spot on).

aaron

aaronbrethorst:
And also, it would be awesome if it used an blurred effects view with vibrance enabled as its background.

tomtclai:
@aaronbrethorst A potentially stupid question. I looked at OBASearchResultsMapViewController.xib and noticed autolayout is not enabled, what do we use to layout buttons?

aaronbrethorst:
@tomtclai Autoresizing masks, aka springs and struts.

tomtclai:
@aaronbrethorst Thanks. How about this:

|          |             |
| ------------- |-------------|
|img_5741 | img_5742|

https://youtu.be/MbxocF336Ic

Some problems I bumped into:
 - looks like Travis Cl check failed because I tried to bold text with [UIFont systemFontOfSize:weight:], avail iOS 8.2. Might need a different way to show the currently selected map
 - I would like to add vibrancy to the PNG as well
 - text is hard to see on a light background (standard maps)
 - minor UI bug in animation 
screenshot 2015-08-24 20 34 05

aaronbrethorst:
@tomtclai this is looking cool!

- I agree that the text is hard to read on a standard map. How about trying the dark variant?
- We'll need new icons for standard and hybrid. I'll see if I can get that resolved.
- The next version of OBA requires 8.0.

tomtclai:
Using the dark variant across the board is easy. But it doesn't look amazing with satellite maps. I'm working on only using the dark variant on the standard maps.

|          |             |
| ------------- |-------------|
| img_5746 | img_5743 |
| img_5748 | img_5747

aaronbrethorst:
OK, let's switch between them on different views.

tomtclai:
https://youtu.be/Vf3DJB4auZA

tomtclai:
I looked at Google maps and see that they only have an on/off option for either hybrid or standard map. Do we really need pure satellite maps? It could be one button with a filled/non-filled satellite icon and that would make things simpler.

Hybrid | Standard
----------| --------
img_5857|img_5855

PR #441, opened 2015-09-13
fixes #440, as well as warnings about deployment target inconsistency

### LINKED ISSUE #440: Project does not build with Xcode 7 GM. can't find the .xcodeproj to disable Bitcode.
>ld: '/Users/tomlai/onebusaway-iphone/Pods/GoogleAnalytics-iOS-SDK/libGoogleAnalyticsServices.a(TAGDataProvider.o)' does not contain bitcode. You must rebuild it with bitcode enabled (Xcode setting ENABLE_BITCODE), obtain an updated library from the vendor, or disable bitcode for this target. for architecture arm64
>clang: error: linker command failed with exit code 1 (use -v to see invocation)

I know i can [disable Bitcode](http://stackoverflow.com/a/30856364/3614291) if I can find the .xcodeproj for GoogleAnalytics, but looks like it doesn't exist. I tried disabling Bitcode in all the .xcodeproj files I could find (there are 3, right?), but no luck.

Thoughts?

tomtclai:
The deployment target should be 8.0, right?


aaronbrethorst:
awesome :+1:

PR #447, opened 2015-10-28
Revert "Display an alert when we fail to retrieve data from the server."

aaronbrethorst:
Reverts OneBusAway/onebusaway-iphone#446


aaronbrethorst:
(self-flagellates)

PR #455, opened 2015-11-01
Enable deep-linking into the Settings app to let users re-enable location services

aaronbrethorst:
* Show an alert when location services are disabled for the app
* Create a utility class to store that alert, and future global alerts
* Remove OBALocationManager's `disabled` property. Rely solely upon `CLLocationManager` for that information.
* Always keep the map view's left location button enabled. Show the "Fix It" deep linking dialog every time it's tapped when location services are disabled.


bbodenmiller:
Not sure how to address this but if you dismiss the prompt to fix location services and then click the find location button (top left) it does not work and no further prompts are displayed.

PR #473, opened 2015-12-04
Remove pods

aaronbrethorst:
* Add Pods folder to gitignore
* Delete the Pods folder
* Regenerate the workspace


bbodenmiller:
@aaronbrethorst do we need to update the readme and tell people to run:

```
gem install cocoapods
pod install
```

Or whatever the correct commands are?

PR #474, opened 2015-12-04
update .gitignore

bbodenmiller:
with latest details from https://github.com/github/gitignore/blob/master/Objective-C.gitignore


aaronbrethorst:
Thought you were using a Nexus 5X... :)

PR #477, opened 2015-12-06
Fix #347

### LINKED ISSUE #347: Warn about invalid custom APIs
As mentioned at https://github.com/OneBusAway/onebusaway-android/issues/105#issuecomment-71123903 many people are using the custom API field as a freetext field for things like `seattle wa`. Perhaps validating the API and warning them if it is invalid would be useful.

aaronbrethorst:
Fixes #347

* Add tests to validate new helper method
* Create a new helper method to create valid URLs out of whatever junk users type in
* Validate that the URLs typed in point to OBA servers
* Add SVProgressHUD to indicate async operations
* Update credits.html with reference to SVProgressHUD
* Improve how -refreshSettings determines its correct server URL


barbeau:
:+1: I agree using current-time is a good way to validate the endpoint.

Issue #484, opened 2015-12-07
Use SFSafariViewController instead of `-[UIApplication openURL:]`


aaronbrethorst:
Once we drop support for iOS 8, we can safely use `SFSafariViewController` everywhere where we currently use a web view to display remote content in the app.

Proper resolution for this task:

* replace all calls to `-[UIApplication openURL:]` with modally-presented `SFSafariViewController`s
* Display StopInfo pages inside of `SFSafariViewController`, too

PR #487, opened 2015-12-08
use SFSafariViewController as well as WKWebView for the credits

tomtclai:
#484


aaronbrethorst:
thanks, but this needs to wait until we ditch iOS 8, which is why it was placed in the 2.5 milestone.

tomtclai:
Right, Should I close this for now?

PR #511, opened 2015-12-17
Simplify the travis.yml file

aaronbrethorst:
I'm pretty sure we can get away with not having a bunch of the stuff that was in our travis.yml file in our travis.yml file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,11 @@
-osx_image: xcode7.1
+osx_image: xcode7.2
 language: objective-c
-before_install:
-  - brew update
-  - brew outdated xctool || brew upgrade xctool
-  - gem install cocoapods
-  - pod install
 xcode_workspace: org.onebusaway.iphone.xcworkspace
-xcode_sdk:
-    - iphonesimulator
 xcode_scheme:
     - Debug
 notifications:
   slack:
-    rooms: 
+    rooms:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,18 +1,11 @@
-osx_image: xcode7.1
+osx_image: xcode7.2
 language: objective-c
-before_install:
-  - brew update
-  - brew outdated xctool || brew upgrade xctool
```
bbodenmiller: This was just to make sure the latest version of `xctool` was installed. Before the Travis images weren't keeping up with the updates very frequently.

.travis.yml
```diff
@@ -1,18 +1,11 @@
...
-  - gem install cocoapods
-  - pod install
```
bbodenmiller: I think you need this to install the cocoapods?

.travis.yml
aaronbrethorst: ahh gotcha, thanks for the info.

.travis.yml
bbodenmiller: Actually looks like this may run automatically now: https://docs.travis-ci.com/user/languages/objective-c/#Dependency-Management

aaronbrethorst: the docs seem to imply that they do this implicitly if you have a Podfile checked in. I thought I'd test it in the best way I know how ;-)

Issue #514, opened 2015-12-18
Some stops disable the "Tap to load more arrivals" button after one load


aaronbrethorst:
From a user:

> I did see one strange anomaly with the ‘Tap to load more arrivals’ function. It works as expected at most stops, but at stops #1120 - W bound, #590 - NW bound and other stops located on 3rd Avenue, the app will only allow for one use of the ‘Tap to load more arrivals.’ After that, the link is ghosted. This is not the case at other stops I use.

I've been able to reproduce the issue at stop 1120. The logic in `OBAGenericStopViewController` for figuring out what's enabled and disabled, and what goes where is a bit crazy. It would be awesome to overhaul this and rebuild the logic for figuring out which cells go where on top of my static table code.

yogipatel:
On it.

aaronbrethorst:
fixed w/ the MVVM rearchitecture.

PR #519, opened 2016-01-14
Fixes #516. Removes duplicate stop header when filtering.

### LINKED ISSUE #516: Duplicate stop info headers
image1

yogipatel:
The logic for sections/headers/rows of OBAGenericStopViewController is quite convoluted. This should serve as a stop-gap for now, but should definitely be cleaned up at some point in the near future. That would also address #514.

### DIFF
--- a/ui/stop_details/OBAGenericStopViewController.m
+++ b/ui/stop_details/OBAGenericStopViewController.m
@@ -475,7 +475,12 @@ static NSString *kOBANoStopInformationURL = @"http://stopinfo.pugetsound.onebusa
 }
 
 - (CGFloat)tableView:(UITableView *)tableView heightForHeaderInSection:(NSInteger)section {
-    return kSectionHeaderHeight;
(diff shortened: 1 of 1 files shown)


ui/stop_details/OBAGenericStopViewController.m
```diff
@@ -475,7 +475,12 @@ - (BOOL)sectionNeedsExplanatoryFooterMessage:(NSInteger)section {
 }
 
 - (CGFloat)tableView:(UITableView *)tableView heightForHeaderInSection:(NSInteger)section {
-    return kSectionHeaderHeight;
+    if ([_filteredArrivals count] != [_allArrivals count] && section == 1) {
```
aaronbrethorst: please use properties and dot notation whenever possible, e.g. `self.filteredArrivals.count`

Also, instead of checking the raw section index, please check `-sectionTypeForSection:` to make sure that it's the expected section type.

aaronbrethorst:
Looks :+1: modulo the feedback above

PR #522, opened 2016-01-16
Cleaned up OBAApplicationDelegate + OBAStopIconFactory

yogipatel:
Removed OBAApplicationDelegate from being handed-off to every view controller, as many didn't need it. Made OBAStopIconFactory 100% class methods since the class carried no state.


aaronbrethorst:
nice!

PR #547, opened 2016-03-12
Overhaul UI for filtering departures

aaronbrethorst:
Move it to a segmented control toggle instead of a table cell that you tap.

* New static cell type: segmented control
* Plumb the segmented control into the Stop UI

### DIFF
--- a/org.onebusaway.iphone.xcodeproj/project.pbxproj
+++ b/org.onebusaway.iphone.xcodeproj/project.pbxproj
@@ -20,6 +20,8 @@
 		932076471C90CE12005F7D4A /* bookmark_without_region_identifier.plist in Resources */ = {isa = PBXBuildFile; fileRef = 932076461C90CE12005F7D4A /* bookmark_without_region_identifier.plist */; };
 		932076491C90CF73005F7D4A /* bookmark_with_region_identifier.plist in Resources */ = {isa = PBXBuildFile; fileRef = 932076481C90CF73005F7D4A /* bookmark_with_region_identifier.plist */; };
 		9320764C1C915441005F7D4A /* OBABookmarksViewController2.m in Sources */ = {isa = PBXBuildFile; fileRef = 9320764B1C915441005F7D4A /* OBABookmarksViewController2.m */; };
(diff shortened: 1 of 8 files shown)


barbeau:
What does this look like on iOS?  I've been thinking about implementing on Android.

aaronbrethorst:
@barbeau OLD:

img_1140

aaronbrethorst:
@barbeau NEW:

simulator screen shot mar 12 2016 9 40 51 am

barbeau:
Thanks!

barbeau:
BTW - how are you doing the walk to stop distance/time?  Google Directions API?

aaronbrethorst:
@barbeau stop distance and time are provided by Apple as part of their MapKit API. Here's some info about it: http://initwithfunk.com/blog/2014/02/14/you-dont-need-to-pay-google-to-get-directions-and-geocoding/

barbeau:
@aaronbrethorst Ok, thanks.

And sorry, another question - what rules exactly are being used to filter out departures by default?

aaronbrethorst:
@barbeau - there are no rules. it is entirely driven by the user. There's a "Filter & Sort Routes" option at the bottom of the view where the user can exclude routes that they don't want to see. By default, every route is displayed. This toggle only appears if the user has filtered at least one route.

barbeau:
Ah, I see - so this option isn't to filter arrivals vs. departures, but just to filter out arrivals **AND** departures for certain routes.  In that case we do support this route filtering on Android.

On Android I'm trying to solve the problem of better differentiating arrival vs. departure info (general issue at https://github.com/OneBusAway/onebusaway-android/issues/287), and in the case of loop routes automatically filter out one of them (https://github.com/OneBusAway/onebusaway-android/issues/436).  The loop route issue is the more confusing and immediate issue I'm working on resolving.  I assume this also exists on iOS, then, although I don't think we have an issue open for it.  I'll open one shortly.

barbeau:
New issue opened specifically for loop routes at https://github.com/OneBusAway/onebusaway-iphone/issues/562.

PR #560, opened 2016-03-14
Fix #62 - Show stop icons at a higher zoom level

### LINKED ISSUE #62: Show stop icons at a higher (i.e., more zoomed out) zoom level
Originally referenced in #42.

The apps cuts off the view of icons at too low a zoom level.  Right now I'm several blocks from transit service that I wanted to look at, which requires me to zoom out, pan elsewhere, and then zoom back in to get stop icons to appear on the map.  And if I zoom out by a step or so, the icons disappear and again prompt me to "Zoom in to look for stops".  

In contrast, in the Android app I can zoom out to almost the city level and still get icons for all the stops on the map.

We should allow stop icons to be shown at least a few zoom levels higher than the current cutoff that triggers the "Zoom in to look for stops" prompt.

cagryInside:
I bump the zoom level.

The furthest zoom level looks like this in the app:

screen shot 2016-03-14 at 11 28 13 am

cc'd @barbeau

### DIFF
--- a/OBAKit/Helpers/OBAMapHelpers.m
+++ b/OBAKit/Helpers/OBAMapHelpers.m
@@ -17,7 +17,7 @@
 const double OBADefaultMapRadiusInMeters = 100;
 const double OBAMinMapRadiusInMeters = 150;
 //const double OBAMaxLatitudeDeltaToShowStops = 0.008;
-const double OBAMaxLatitudeDeltaToShowStops = 0.01;
+const double OBAMaxLatitudeDeltaToShowStops = 0.05;
 const double OBARegionScaleFactor = 1.5;
 const double OBAMaxMapDistanceFromCurrentLocationForNearby = 800;
 const double OBAPaddingScaleFactor = 1.075;


barbeau:
@aaronbrethorst Could you please take a look and see if you're ok with this zoom level?  This is the zoom level PSTA is expecting to see stops at.

aaronbrethorst:
@cagryInside is this sufficient for Tampa's needs?

barbeau:
@aaronbrethorst Crossing messages :) - yes, this should be sufficient for Tampa.

aaronbrethorst:
@barbeau @cagryInside great! this is totally fine from my perspective. the amount of data being used is still far less than what's required to send map tiles over the wire. i think this'll be a big win in terms of overall ux.

aaronbrethorst:
@cagryInside Also, for future reference, you can press Command S in the Simulator to capture _only_ the Simulator's content with none of the window frame.

aaronbrethorst:
:+1:

barbeau:
:+1:  Thanks!

cagryInside:
@aaronbrethorst This is the simulator screenshot:

simulator screen shot mar 14 2016 1 02 01 pm

PR #568, opened 2016-03-15
Fix #565 -  Refresh stops rendering when zoom level change

### LINKED ISSUE #565: Stop rendering is inconsistent when zooming in and out
### bug report from a user

Aaron,

I found a bug with zooming in and out of the map, specifically as it relates to the feature you describe where more stops are visible when you zoom out. If I am zoomed in on the way, I see all of the stops in my neighborhood. But if I zoom out, several stops disappear. It appears that it is perhaps the desired UI behavior for clarity, otherwise the entire map would just be a sea of stops overlaid on the map. But when I zoom back into the neighborhood and street level, most of my local stops do not return to the UI. I only see the select few stops that were there when I was zoomed out. Does this make sense?

### App Version/Hardware

March 14 TestFlight build.

cagryInside:
Added a zoom level control that updates stops when the zoom level changes:

```
if (!moreAccurateRegion && containedRegion && !zoomLevelChanged) {
          NSString *label = [self computeLabelForCurrentResults];
          [self applyMapLabelWithText:label];
          return;  // Do NOT update stops
}
```
cc'd @aaronbrethorst @barbeau

### DIFF
--- a/OBAKit/Helpers/OBAMapHelpers.h
+++ b/OBAKit/Helpers/OBAMapHelpers.h
@@ -15,6 +15,7 @@ extern const double OBAMaxLatitudeDeltaToShowStops;
 extern const double OBARegionScaleFactor;
 extern const double OBAMaxMapDistanceFromCurrentLocationForNearby;
 extern const double OBAPaddingScaleFactor;
+extern const double OBARegionZoomLevelThreshold;
 
 @class OBAPlacemark;
 @class OBARegionBoundsV2;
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
neat, i'll load this onto my phone and use it for the commute home. I'll take a real peek at the code this evening once I'm home.

aaronbrethorst:
looks good!

cagryInside:
@aaronbrethorst thanks

barbeau:
Thanks @cagryInside!

PR #569, opened 2016-03-17
Show Bookmarks on the Map

### LINKED ISSUE #3: Show bookmarked stops at any zoom level (with a different icon?)
It might be useful if bookmarked stops were shown at any zoom level, maybe even with a different icon.

 Bookmarked stops could also be shown when zoomed out on a route map.

aaronbrethorst:
Fixes #3 - Show bookmarked stops at any zoom level (with a different icon?) (https://github.com/OneBusAway/onebusaway-iphone/issues/3)

Map Improvements:

* Remove the "result exceeded" error message on the map view, which isn't super helpful.
* Add tinted stop icons to the map for bookmarks, and show them at any zoom level.
* Rename some of the local variables around annotations in the map search results controller in order to improve clarity

Stops and Bookmarks:

* Rename the OBABookmarkV2 property "stopID" to "stopId" so we can take advantage of -respondsToSelector and id types for determining stopIds of stop views to push onto the stack.
* Add route type to OBABookmarkV2
* Make it easier to access route type from a bookmark
* Store OBAStopV2 inside of an OBABookmarkV2 to make it easier to get to necessary data from the map
* Update the stop controller so that it injects the correct stop into bookmarks that don't have one assigned yet.

Misc:

* Add separate file for OBARouteType enum


barbeau:
@aaronbrethorst This is cool!  Do you have a screenshot you can share for how the favorites look different from the normal stops?

aaronbrethorst:
@barbeau sure

image

PR #626, opened 2016-05-20
Fix #624 - Add analytics to the service alerts button

### LINKED ISSUE #624: Add analytics to the "service alerts button"
**Summary:** 

We would like to report how many times a service alert button is clicked. Therefore, we need to add analytics when a service alert button is clicked. We should also include the agency id to this report.

cagryInside:
Following analytics report was added when a `service alerts button` is clicked:

 Event Category  | Event Action | Event Label |
| ------------- | ------------- | ------------- |
| ui_action  | service_alerts  | Clicked Service Alerts from: PSTA |

Screenshot from GA console:

untitled 2

cc'd @barbeau

### DIFF
--- a/OBAKit/Models/unmanaged/OBAReferencesV2.h
+++ b/OBAKit/Models/unmanaged/OBAReferencesV2.h
@@ -17,6 +17,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void) addAgency:(OBAAgencyV2*)agency;
 - (OBAAgencyV2*) getAgencyForId:(NSString*)agencyId;
+- (NSDictionary*) getAllAgencies;
 
 - (void) addRoute:(OBARouteV2*)route;
 - (OBARouteV2*) getRouteForId:(NSString*)routeId;
--- a/OBAKit/Models/unmanaged/OBAReferencesV2.m
+++ b/OBAKit/Models/unmanaged/OBAReferencesV2.m
@@ -27,7 +27,11 @@
 
 - (OBAAgencyV2*) getAgencyForId:(NSString*)agencyId {
     return _agencies[agencyId];
-}    
+}
+
+- (NSDictionary*) getAllAgencies {
+    return [NSDictionary dictionaryWithDictionary:_agencies];;
+}
 
 - (void) addRoute:(OBARouteV2*)route {
     _routes[route.routeId] = route;
--- a/ui/situations/OBASituationsViewController.m
+++ b/ui/situations/OBASituationsViewController.m
@@ -12,6 +12,7 @@
 #import "UITableViewController+oba_Additions.h"
 #import "UITableViewCell+oba_Additions.h"
 #import "OBAAnalytics.h"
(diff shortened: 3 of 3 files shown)


barbeau:
Thanks @cagryInside!  

@aaronbrethorst Would you be able to get this out in the next release?

bbodenmiller:
I don't think you need to specify the region as I believe you can determine that by cross referencing the data.

cagryInside:
> I don't think you need to specify the region as I believe you can determine that by cross referencing the data.

@bbodenmiller To clarify, it is not region name, it is `agency_id`. Since, a region can contain multiple agencies, this helps us to differentiate service alerts from different agencies in the same region.

OBAKit/Models/unmanaged/OBAReferencesV2.h
```diff
@@ -17,6 +17,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void) addAgency:(OBAAgencyV2*)agency;
 - (OBAAgencyV2*) getAgencyForId:(NSString*)agencyId;
+- (NSMutableDictionary*) getAllAgencies;
```
aaronbrethorst: This class shouldn't publicly vend a mutable dictionary. I'd like to see the method signature changed to `NSDictionary`

OBAKit/Models/unmanaged/OBAReferencesV2.m
```diff
@@ -27,7 +27,11 @@ - (void) addAgency:(OBAAgencyV2*)agency {
 
 - (OBAAgencyV2*) getAgencyForId:(NSString*)agencyId {
     return _agencies[agencyId];
-}    
+}
+
+- (NSMutableDictionary*) getAllAgencies {
+    return _agencies;
```
aaronbrethorst: `return [NSDictionary dictionaryWithDictionary:_agencies];`

ui/situations/OBASituationsViewController.m
```diff
@@ -93,6 +94,17 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
...
         [self.navigationController pushViewController:vc animated:YES];
+        [self reportAnalytics:situation];
+    }
+}
+
+- (void) reportAnalytics:(OBASituationV2 *) situation {
+    
+    NSMutableDictionary *agencies = [[situation references] getAllAgencies];
+    for(id key in agencies) {
+         OBAAgencyV2 * agency = [agencies objectForKey:key];
```
aaronbrethorst: please use object literal syntax: `agencies[key]`

aaronbrethorst:
please make the changes described above and squash the resulting commits, and then i'll be happy to merge. thank you!

cagryInside:
> please make the changes described above and squash the resulting commits, and then i'll be happy to merge. thank you!

@aaronbrethorst thanks for your comments. Latest commit fixes the in-line comments.

aaronbrethorst:
looks good!

PR #630, opened 2016-05-29
Update readme for cocoapods compatibility info.

chadsy:
...because I just learned the hard way 👍


aaronbrethorst:
What is the incompatibility?

chadsy:
```
[!] Invalid `Podfile` file: [!] The specification of `link_with` in the Podfile is now unsupported, please use target blocks instead..
```

aaronbrethorst:
Thanks, chad :)

PR #670, opened 2016-08-27
When "cancel" is pressed reset the navigationItem title to "Map"

### LINKED ISSUE #621: Back button says "Agencies" instead of "Map" in edge case
### Expected behavior
When viewing a stop from the map view, I expect the back button to say < Map

### Actual behavior
When viewing a stop from the map view, the back button says < Agencies

Note: these repro steps only work after the crash when viewing agencies on a map is fixed (aka they don't work on the current production app 2.5.2)
1. Go to the Info tab
2. Tap "Agencies"
3. Tap "View on Map" at the top
4. Tap the "Cancel" button to get out of the Agencies view. This leaves you in the map view, but removes all the agency markers
5. Zoom in if necessary and tap on any stop and tap the *i* icon to view details

Note: switching tabs at the bottom and going back to the map tab and then viewing details on a stop does not fix the issue

simulator screen shot apr 22 2016 12 10 07 am

### App Version/Hardware
iPhone 6 Simulator, develop branch with fix for #619 applied

RPangrle:
Fixes #621

### DIFF
--- a/ui/search/OBASearchResultsMapViewController.m
+++ b/ui/search/OBASearchResultsMapViewController.m
@@ -1144,6 +1144,7 @@ static const double kStopsInRegionRefreshDelayOnDrag = 0.1;
 }
 
 - (void)cancelPressed {
+    self.navigationItem.title = NSLocalizedString(@"Map", @"Map tab title");
     self.navigationItem.titleView = self.titleView;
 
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
👍

PR #720, opened 2016-09-12
Ensure that bookmarks have titles before allowing them to be created

### LINKED ISSUE #711: Bookmarks should be validated as having non-zero length titles before being created
### Expected behavior

A bookmark without a name cannot be created.

### Actual behavior

It can be created, which causes other problems later on.

Create a bookmark, leave the name blank.

### App Version/Hardware

2.6.0

aaronbrethorst:
* Add a -validateModel method to OBABookmarkV2 that checks to see if the bookmark has a name with length greater than 0
* Add tests demonstrating that -validateModel works correctly
* Show an AlertPresenter-based error message if the -validateModel check fails

Fixes #711 - Bookmarks should be validated as having non-zero length titles before being created

### DIFF
--- a/OBAKit/Models/unmanaged/OBABookmarkV2.h
+++ b/OBAKit/Models/unmanaged/OBABookmarkV2.h
@@ -39,6 +39,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithStop:(OBAStopV2*)stop region:(OBARegionV2*)region;
 
 /**
+ Tests whether this is a valid bookmark model.
+
+ @return True if it is valid and false if it is invalid.
+ */
+- (BOOL)isValidModel;
+
+/**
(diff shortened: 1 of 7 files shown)


OBAKit/Models/unmanaged/OBABookmarkV2.h
```diff
@@ -39,6 +39,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithStop:(OBAStopV2*)stop region:(OBARegionV2*)region;
 
 /**
+ Tests whether this is a valid bookmark model.
+
+ @return True if it is valid and false if it is invalid.
+ */
+- (BOOL)validateModel;
```
chadsy: `isValidModel` perhaps?

chadsy:
👍 % your thoughts on the comment.

PR #721, opened 2016-09-12
Enabled aggressive build and analyzer warnings.

aaronbrethorst:
* Turned on "treat warnings as errors"
* Promoted some build settings from target to project level
* Deleted redundant target-specific settings
* Turned on a raft of aggressive build warnings, fixed the few issues they turned up
* Fixed a handful of static analyzer issues, including some nullability warnings
* Added a few TODOs for nullability, where we're not conforming to a platform/3rd party protocol


chadsy:
👍 Thanks for the help and pull thru on this!

PR #722, opened 2016-09-12
Fixes #717 - Don't allow saving of empty Group name

### LINKED ISSUE #717: Bookmark Groups should be validated as having non-zero length titles before being created
### Expected behavior

A bookmark group without a name cannot be created.

### Actual behavior

It can be created, which causes other problems later on.

Create a bookmark group, leave the name blank.

### App Version/Hardware

2.6.0

chadsy:
* Elected to fix this by enabling/disabling the Save button based on text field length
* Also collapsed redundant UIAlertController code into one method for add/edit


chadsy:
@aaronbrethorst if you're not a fan of this solution, i can duplicate the fix for the bookmark names (show an error).

aaronbrethorst:
I think you're going to have to implement this on the model and check that instead. There's a whole other experience that's accessible via Stop -> Add Bookmark -> Select Route -> Group -> Add. Both should have the same validation for their models. lmk if you have any questions.

chadsy:
Thanks for pointing out the additional case. Added some discussion to the bug, will proceed as described there and update this PR.

aaronbrethorst:
@chadsy don't forget you need to git fetch upstream, then rebase against upstream/develop, and then push -f

ui/bookmark_groups/OBABookmarkGroupsViewController.m
```diff
@@ -58,24 +58,53 @@ - (void)close {
     [self dismissViewControllerAnimated:YES completion:nil];
 }
 
-- (void)addGroup {
-    UIAlertController *alertController = [UIAlertController alertControllerWithTitle:NSLocalizedString(@"Add Bookmark Group", @"") message:nil preferredStyle:UIAlertControllerStyleAlert];
+static UIAlertAction *saveButton;
```
aaronbrethorst: Why not make this a property on the object? If you want to keep this as is, please set to nil initially.

ui/bookmark_groups/OBABookmarkGroupsViewController.m
```diff
@@ -58,24 +58,53 @@ - (void)close {
...
+
+- (void)textFieldDidChange :(UITextField *)textField {
```
aaronbrethorst: Wonky spacing around param

aaronbrethorst:
looks great, thank you! 🎆

PR #729, opened 2016-09-17
Don't crash if the arrival and departure object lacks a headsign

### LINKED ISSUE #723: Ensure the app properly handles trips with null headsigns
See https://github.com/OneBusAway/onebusaway-android/issues/690 for more info. Make sure that OBA iOS gracefully handles this scenario.

aaronbrethorst:
Fixes #723

---

Also, remove an unnecessary property from a view and turn it into a local variable instead.


chadsy:
👍

PR #741, opened 2016-09-18
Only allow the user to see vehicle details if a vehicle exists

### LINKED ISSUE #735: Attempting to view vehicle info on a scheduled route shows "Error connecting"
### Expected behavior
Either no ability to look at Vehicle info if it's not available, or a clearer message that vehicle info isn't available with scheduled rather than actual data.

### Actual behavior
Vehicle Info screen appears, showing a title of either "Error connecting" or "Unknown error" and "Updating..." in the view.

1. Open a stop with several route arrivals
1. Choose an arrival showing "scheduled arrival" instead of an actual live time
1. Tap Vehicle Info at the bottom of the route list screen
1. Notice the Vehicle Info screen...

image

aaronbrethorst:
Fixes #735 - Attempting to view vehicle info on a scheduled route shows "Error connecting"

* If the vehicle ID attached to the trip status that the user is looking at is blank, then show a "No vehicle info available" message
* Delete a couple unused methods that I noticed


aaronbrethorst:
@chadsy please take a peek when you have a sec and let me know what you think about this fix. https://github.com/OneBusAway/onebusaway-iphone/pull/741/files#diff-1e93d4f68f7dbe77de60722732435f68R230

chadsy:
🚢 🇮🇹

PR #743, opened 2016-09-18
Show explanatory message about scheduled arrivals on Bookmarks tab

### LINKED ISSUE #686: The bookmarks tab should explain what an '*' means on a scheduled route
Recommendation: Put a table footer view into the table along the lines of what's being done in the `OBABookmarkGroupsViewController`. Note the comment in that view controller from lines 33-41, though, which will explain an additional feature that this tab's footer will have to include.

Bonus points: figure out how to abstract out the preservation of the original footer view so that this long-winded comment and fiddly logic doesn't have to be strewn all over the app.

-----

### Expected behavior

The bookmarks tab includes a footer that says '* means that this is a scheduled, etc etc. No real time data, etc.'

### Actual behavior

It doesn't. It only shows the asterisk.

Bookmark a stop with a route, link the Seattle link light rail, and go to the bookmarks tab.

img_1090

### App Version/Hardware

2.6.0

aaronbrethorst:
Fixes #686 - The bookmarks tab should explain what an '*' means on a scheduled route

* Add 'scheduled departure' string to OBAStrings and use it on bookmarks and stops view controllers
* Show a 'scheduled departure' footer on the bookmarks view controller when one or more of its departures lacks real time data
* Make it easier to add a footer to a table view when DZNEmptyDataSet is being used
* Extract a helper method for determining if an ArrivalAndDepartureForStop object lacks real time data
* Find another possible source of the "non-modular includes" errors and (fingers crossed) maybe actually fix it this time


chadsy:
👍

PR #748, opened 2016-09-19
Always refresh app state when the region changes

### LINKED ISSUE #746: Always refresh settings after changing region
Currently, there's a two step process performed whenever a new region is selected:

1. Set the new region on the model DAO object.
2. Call `-[OBAApplication refreshSettings]`

There's no reason to perform step 2. It's error prone and can be performed automatically now that `OBARegionDidUpdateNotification` is posted whenever the current region changes.

aaronbrethorst:
Fixes #746 - Always refresh settings after changing region


chadsy:
👍

PR #757, opened 2016-09-23
Eliminate unnecessary classic departure row

aaronbrethorst:
The 'static table' MVVM system lends itself extremely well to mapping a view model object (a row) to any number of different representations (a cell). To that end, I've removed the redundant and confusing OBAClassicDepartureRow class. If you want to display an OBAClassicDepartureCell, just set the row's cellReuseIdentifier to the appropriate value and you're off to the races!


chadsy:
👍

PR #758, opened 2016-09-24
Allow 1-3 departure times to appear in one cell

aaronbrethorst:
Use this new rendering style on the bookmarks controller

Also:

* Add NSArray category method to group contents of an array based on a key
* Remove the non-classic departure cell.
* Remove stop sorting code


aaronbrethorst:
simulator screen shot sep 23 2016 9 11 46 pm

aaronbrethorst:
Things to improve:

- [ ] Spacing is ugly
- [ ] Move the "m" below the text and rename it "min"
- [ ] Consider making it clear which of those three routes are on time, delayed, or scheduled arrival

(maybe not in this PR—in which case this should be moved into a new issue)

PR #790, opened 2016-10-07
Enhance Apptentive Integration

### LINKED ISSUE #732: Enhance Apptentive Integration
- [x] Release notes
- [x] Love it/needs work rating prompt
- [x] Improve indication of responses

aaronbrethorst:
Fixes #732 - Enhance Apptentive Integration

* Add accessoryView to OBATableRow
* Only show the Apptentive table row on Info if the integration is up and running
* Show an unread message badge on the Info tab when a message is awaiting the user

Apptentive Events:

* App became active event
* Send an event 10 seconds after the user views a stop

### DIFF
--- a/ui/stops/OBAStopViewController.m
+++ b/ui/stops/OBAStopViewController.m
@@ -26,6 +26,7 @@
 #import "OBAArrivalAndDepartureViewController.h"
 #import "OBAStaticTableViewController+Builders.h"
 #import "OBABookmarkRouteDisambiguationViewController.h"
+#import "Apptentive.h"
 
 static NSTimeInterval const kRefreshTimeInterval = 30.0;
 static CGFloat const kTableHeaderHeight = 150.f;
@@ -38,6 +39,7 @@ static CGFloat const kTableHeaderHeight = 150.f;
 @property(nonatomic,strong) OBAStopPreferencesV2 *stopPreferences;
 @property(nonatomic,strong) OBARouteFilter *routeFilter;
 @property(nonatomic,strong) OBAParallaxTableHeaderView *parallaxHeaderView;
+@property(nonatomic,strong) NSTimer *apptentiveTimer;
 @end
 
 @implementation OBAStopViewController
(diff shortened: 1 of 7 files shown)


ui/stops/OBAStopViewController.m
```diff
@@ -69,6 +85,7 @@ - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
     
     self.refreshTimer = [NSTimer scheduledTimerWithTimeInterval:kRefreshTimeInterval target:self selector:@selector(reloadData:) userInfo:nil repeats:YES];
+    self.apptentiveTimer = [NSTimer scheduledTimerWithTimeInterval:10 target:self selector:@selector(recordUserVisit:) userInfo:nil repeats:NO];
```
chadsy: Seems like this would be cleaner with a `performSelector...afterDelay` or a `dispatch_after` instead of a one-shot timer.

chadsy:
👍 mod comments

OBAKit/Helpers/OBACommonV1.h
```diff
@@ -16,6 +16,9 @@
 
 @import Foundation;
 
+#ifndef _OBA_COMMON_V1_
+#define _OBA_COMMON_V1_
+
```
aaronbrethorst: I didn't think they were either, but the compiler started complaining about double inclusion of this file and I added these guards in a fit of annoyance.

ui/stops/OBAStopViewController.m
aaronbrethorst: True! But what if the user exits the view controller in the intervening time? I only want to display a message to the user if they are still on the stop controller. This raises the very good point that I should probably document this :)

ui/themes/OBAClassicApplicationUI.m
```diff
@@ -55,11 +57,29 @@ - (instancetype)init {
...
         _tabBarController.delegate = self;
+
+        [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(refreshMessageBadge:) name:ApptentiveMessageCenterUnreadCountChangedNotification object:nil];
     }
 
     return self;
 }
 
+- (void)dealloc {
+    [[NSNotificationCenter defaultCenter] removeObserver:self name:ApptentiveMessageCenterUnreadCountChangedNotification object:nil];
```
aaronbrethorst: wow! this is amazing! http://useyourloaf.com/blog/unregistering-nsnotificationcenter-observers-in-ios-9/

PR #791, opened 2016-10-07
Fix issue where Stop Controller refresh buttons 'pops' on reload

### LINKED ISSUE #788: Stop controller 'pops' on data refresh
### Expected behavior

Unless the user pulls to refresh, the stop UI doesn't move around when it loads

### Actual behavior

There's a slight 'pop' when the refresh control is displayed 

1. Open a stop
2. Tap the reload button in the top right 

### App Version/Hardware

2.6.1

aaronbrethorst:
Fixes #788

Only animate reloading if -reloadData: is called from an object other than the right navigation item button.


chadsy:
🐑 🇮🇹

PR #793, opened 2016-10-07
Properly escape forward slashes ('/') that appear in stop IDs

### LINKED ISSUE #787: Error when tapping on stop that contains a forward slash "/" in stopId
**Summary:** 

If you tap on a stop that has a `/` in the `stopId`, the app shows a dialog "Error - The data couldn't be read because it isn't in the correct format".

Opened similar issue on OBA Android:
https://github.com/OneBusAway/onebusaway-android/issues/704

Seems like client-side we should probably escape `/` as `%2F` - see http://stackoverflow.com/questions/2992231/slashes-in-url-variables. 

Just opened server issue here as well - it's not clear that even if we escape the `/` client-side that servers can accept this in a REST API paramater :
https://github.com/OneBusAway/onebusaway-application-modules/issues/182

**Steps to reproduce:** 

1. Using Boston region, search for stop `Forge Park`
1. Show stop on map and tap to see arrivals

It has the following stopId:
`1_Forge Park / 495`

...and it can be seen in the list of stops returned in this stops-for-location query:
http://developer.onebusaway.org/mbta-api/api/where/stops-for-location.json?lat=42.078482127189595&lon=-71.43382396548986&latSpan=0.08702002322407765&lonSpan=0.07993753999471664&version=2&key=TEST

**Expected behavior:** 

Show arrival times for that stop

**Observed behavior:** 

App shows dialog "Error - The data couldn't be read because it isn't in the correct format".

**Screenshots:**

image

aaronbrethorst:
Fixes #787 - Error when tapping on stop that contains a forward slash "/" in stopId

Also:

* Add GPX file for Boston
* Make sure the AlertPresenter shows the warning theme when appropriate, as opposed to always showing success

### DIFF
--- a/OBAKit/Services/OBAModelService.m
+++ b/OBAKit/Services/OBAModelService.m
@@ -51,16 +51,18 @@ static const CLLocationAccuracy kRegionalRadius = 40000;
 
 - (AnyPromise*)requestStopForID:(NSString*)stopID minutesBefore:(NSUInteger)minutesBefore minutesAfter:(NSUInteger)minutesAfter {
 
-    OBAGuard(stopID) else {
+    OBAGuard(stopID.length > 0) else {
         return nil;
     }
+
+    NSString *escapedStopID = [stopID stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLHostAllowedCharacterSet]];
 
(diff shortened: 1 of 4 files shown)


chadsy:
👍

PR #799, opened 2016-10-12
Fixes #789 - replace OBAIconFactory and assets with drawing code

### LINKED ISSUE #789: Replace OBAStopIconFactory with something a little more flexible
`OBAStopIconFactory` has served us very well for a very long time, but, as OBA becomes multi-lingual, we will need to look at replacing it. 

I'd like to keep the public interface and the `UIImage`s that are output 100% unchanged for the time being, but I'd like to see the images that are generated be created mostly in code. 

# Example

busstopicone 2x

Let's take the above bus stop icon as an example. The only raster image that should be used to compose this icon should be the icon of the bus itself. The white background, black squircle, arrow, and the word "STOP" should all be rendered in code.

This should work for all supported route types: 

- [ ] Ferry
- [ ] Rail
- [ ] LightRail
- [ ] Bus

and it should work for all directions. 

Icons should be rendered on demand, work for any screen density (1x to 3x), and be cached after creation.

chadsy:
There is some commentary in the comments that describes the geometry of how the icon is constructed.

Basically, here are the highlights:
- constructed a set of UIViews to draw the pieces, then rendered the UIView to a UIImage
- the variables (nee constants) in the drawing, describing percentages, etc., are in a set of #defines at the top of the factory class
- the resulting UIImages are cached with the transport name, direction, and size as the cache key
- in order to draw in a raw view, there is a simple derived class from UIView that calls out to a block in its drawRect, so i could use the same class for each of the special drawing components by providing a different block at init time for each instance
- there are some tiny geometric differences between the static icons and the new code, but these are negligible

Stuff I'm not 100% certain about:
- might want to enhance OBATheme to contain the typeface family and colors used to paint the icons
- The font size calculation (used to determine the largest possible size to fit the "STOP" text into the icon) might have some anomalies with other text, but it works ok for "STOP"


OBAKit/Helpers/OBACanvasView.h
```diff
@@ -0,0 +1,19 @@
...
+//  Copyright © 2016 OneBusAway. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@interface OBACanvasView : UIView
+
+// This is a lightweight view derived from UIView to offer only an opportunity
+// to have a custom drawRect implementation with a reusable class rather than
+// having N tiny classes with different drawing code. Because there's no other
```
aaronbrethorst: neat 👍

OBAKit/Helpers/OBACanvasView.m
```diff
@@ -0,0 +1,37 @@
...
+
+// This is a lightweight view derived from UIView to offer only an opportunity
+// to have a custom drawRect implementation with a reusable class rather than
+// having N tiny classes with different drawing code. Because there's no other
+// functionality needed, I chose a simple block model to do the drawing.
+
+#pragma mark - OBACanvasView
+
+@implementation OBACanvasView {
+    void(^_drawBlock)(CGRect);
```
aaronbrethorst: Any reason this isn't a property? I'm not sure it really makes a difference. Just curious about your decision.

OBAKit/Helpers/OBACanvasView.m
```diff
@@ -0,0 +1,37 @@
...
+    if (self) {
+        _drawBlock = drawBlock;
+        self.backgroundColor = [UIColor clearColor];
+    }
+
+    return self;
+}
+
+- (void)drawRect:(CGRect)rect {
+    _drawBlock(rect);
```
aaronbrethorst: Shouldn't this have a call to `-super`?

OBAKit/Helpers/OBAStopIconFactory.m
```diff
@@ -16,38 +16,298 @@
 
 #import <OBAKit/OBAStopIconFactory.h>
 #import <OBAKit/OBARouteV2.h>
+#import <OBAKit/OBACanvasView.h>
 
 @implementation OBAStopIconFactory
 
+// . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . .
```
aaronbrethorst: :-O you deserve a medal for this comment, really. Seriously, this is amazing.

OBAKit/Helpers/OBAStopIconFactory.m
```diff
@@ -16,38 +16,298 @@
...
+static NSString *stopLabelText;
+static NSCache *iconCache;
+
++(void)initialize {
+    iconCache = [[NSCache alloc] init];
+
+    stopLabelText = NSLocalizedString(@"STOP",);
+
+    // I chose "Helvetica" because it's the official typeface of the MTA
+    stopLabelFaceName = @"Helvetica Bold";
```
aaronbrethorst: I'd prefer that the app uses the system font across the board if possible. It should be possible to easily grab the system typeface from OBATheme

OBAKit/Helpers/OBAStopIconFactory.m
```diff
@@ -16,38 +16,298 @@
...
+
+    // Color selections
+    // TODO: Consider OBATheme here?
+    strokeColor = [UIColor blackColor];
+    iconBackgroundColor = [UIColor whiteColor];
+    textBackgroundColor = strokeColor;
+    chevronFillColor = [UIColor redColor];
+    textColor = [UIColor whiteColor];
+
+    // Rotation transform angles
```
aaronbrethorst: What unit are these? Radians?

OBAKit/Helpers/OBAStopIconFactory.m
```diff
@@ -16,38 +16,298 @@
...
 + (UIImage *)getIconForStop:(OBAStopV2 *)stop {
-    NSString *routeIconType = [self getRouteIconTypeForStop:stop];
-    NSString *direction = @"";
+    return [self getIconForStop:stop withSize:CGSizeMake(StopIconDefaultWidth, StopIconDefaultHeight)];
+}
 
-    if (stop.direction) {
-        direction = stop.direction;
++ (UIImage *)getIconForStop:(OBAStopV2 *)stop withSize:(CGSize)size {
+    // First, let's compose the cache key out of the name and orientation, then
```
aaronbrethorst: Any risk of race conditions?

OBAKit/Helpers/OBAStopIconFactory.m
```diff
@@ -16,38 +16,298 @@
...
+#define ChevronWidthPercentage      0.25f   // Width of chevron, relative to overall size (assuming N chevron)
+#define ChevronHeightPercentage     0.15f   // Height of chevron, relative to overall size (assuming N chevron)
+
+// For the simple getIconForStop call, here are the default dimensions
+#define StopIconDefaultWidth        54
+#define StopIconDefaultHeight       54
+
+// . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . .
+
+static NSDictionary *rotationAngles;
```
aaronbrethorst: Are these guaranteed to be nil by default?

OBAKit/Helpers/OBAStopIconFactory.m
```diff
@@ -16,38 +16,298 @@
...
+                                                                  cornerRadii:CGSizeMake(cornerRadius, cornerRadius)];
+#pragma clang diagnostic pop
+
+        [textBackgroundColor setFill];
+        [rectanglePath fill];
+
+        CGRect labelRect = CGRectInset(blockRect, strokeWidth, strokeWidth);
+        CGFloat fontSize = [self findFontSizeForFrame:labelRect forFace:stopLabelFaceName withText:stopLabelText];
+
+        NSDictionary *attributes = @{ NSFontAttributeName:[UIFont fontWithName:stopLabelFaceName size:fontSize],
```
aaronbrethorst: Should this font be cached too?

OBAKit/Helpers/OBACanvasView.m
chadsy: Mostly because I generally choose ivar over property unless it requires a setter override or changes during the lifetime of the containing instance.

OBAKit/Helpers/OBACanvasView.m
chadsy: Actually...

> "If you subclass UIView directly, your implementation of this method does not need to call super. However, if you are subclassing a different view class, you should call super at some point in your implementation."

[UIView drawRect](https://developer.apple.com/reference/uikit/uiview/1622529-drawrect?language=objc)

OBAKit/Helpers/OBAStopIconFactory.m
chadsy: Yeah, but I'll initialize them to nil explicitly.

OBAKit/Helpers/OBACanvasView.m
aaronbrethorst: TMYK!

OBAKit/Helpers/OBAStopIconFactory.m
chadsy: Ok, `stopLabelFaceName = [OBATheme titleFont].familyName;` it is. Had to move the initializers out of `+(void)initialize` and into a dispatch once block since it calls UIKit code.

OBAKit/Helpers/OBAStopIconFactory.m
chadsy: This is always called from a mapkit delegate method, so should forever be on thread 0. The only mutable state is the image cache, and NSCache is thread safe, so there should be no issue.

OBAKit/Helpers/OBAStopIconFactory.m
chadsy: Yup. Clarified the comment.

OBAKit/Helpers/OBAStopIconFactory.m
chadsy: I feel like there are diminishing returns on caching these, since the enclosing referrer is cache-optimized.

aaronbrethorst: Makes sense

chadsy:
Ok, merged/rebased with code review feedback (initialized statics to nil, fetch typeface name from OBATheme, clarified radians comment)

PR #811, opened 2016-10-17
Ensure that the user always has a selected region

### LINKED ISSUE #810: A region is not selected on first launch after the user authorizes location access when location data is unavailable
### Expected behavior

The user is prompted to select a region after authorizing location services if their location cannot be determined.

### Actual behavior

Right now, nothing happens. The app doesn't choose a region and the user is left staring at a map of the United States.

1. Turn off location simulation
2. Reset the simulator.
3. Run the app
4. Authorize location services

aaronbrethorst:
Fixes #810 - A region is not selected on first launch after the user authorizes location access when location data is unavailable

* Always post a notification from the `OBALocationManager` when it fails to find the user's location so that the region helper prompts the user to pick a region.
* Disable `automaticallySelectRegion` if the location manager was unable to retrieve the user's current location when we also happen to not have a region set yet. This way, the user will be able to tap entries in the region list controller when it appears.
* Remove any knowledge of the user's current location from the Region List controller. It doesn't need to know it.

### DIFF
--- a/OBAKit/Location/OBALocationManager.m
+++ b/OBAKit/Location/OBALocationManager.m
@@ -112,11 +112,10 @@ NSString * const OBALocationErrorUserInfoKey = @"OBALocationErrorUserInfoKey";
 }
 
 - (void)locationManager:(CLLocationManager *)manager didFailWithError:(NSError *)error {
-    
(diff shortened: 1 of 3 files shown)


chadsy:
:shipit:

PR #855, opened 2016-11-04
Deep Links and Shared Trips

### LINKED ISSUE #833: 'Follow My Trip' Feature
Blocked on #840 

I imagine that this would look similar to Uber's "Follow My Trip" feature, except that it would all happen in-app (or in an iMessage extension?), and be predicated on tracking via the sharing of a particular vehicle ID.

This needs a lot more thinking-through to be ready for development, but here's more or less what I had in mind:
- Primary UI is either based on deep linking or an iMessage extension
- The user picks the vehicle they'll be riding and the stop they plan on getting off at
- They can send this information to another person via iMessage
- The other person can monitor the status of the particular vehicle heading towards the stop in question. Ideally we reuse existing UI components, and this new experience is layered on top of the arrivals and departures controller, or something.

### LINKED ISSUE #840: Implement Deep Linking
We should get deep linking support working properly in order to enable other features like #833. 

App delegate:

```
- (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserActivity *)userActivity restorationHandler:(void (^)(NSArray *))restorationHandler {
    NSURL *url = userActivity.webpageURL;
    // handle url
}
```

### LINKED ISSUE #847: Add option to delete individual recent stops
### Expected behavior

I should be able to delete individual recent stops

### Actual behavior

I can only delete all of them.

Go to the Recent tab. Swipe on a row. It should offer to let me delete that row.

aaronbrethorst:
This feature wasn't so much driven by specific user request as it is driven to replace the hacky workarounds I've seen users use to accomplish the same thing.

Here are the key points of this commit:

1. Allows users to share the trip they are taking with another person via text message or any other form of communication vended through UIActivityViewController.
2. Trips are shared via URL. These URLs encode the ID of the region that their trips belong to, which means that they are globally unique. They also do not depend on individual OBA region servers, which means that the iOS app doesn't need region server information hardcoded into it.
3. People with OneBusAway for iOS are taken to the trip itself when they tap on one of these links.
4. People without OBA for iOS are taken first to http://onebusaway.co, and then redirected to the appropriate region server's web page for that trip.

-------

Fixes:

* Fixes #833 - 'Follow My Trip' Feature
* Fixes #840 - Implement Deep Linking
* Fixes #847 - Add option to delete individual recent stops

-------

Deep Links/Shared Trips:

* Deep link router class
* Create OBATripDeepLink model object
* Shared Trip infrastructure for model DAO and the user defaults-based persistence layer
* Display shared trips on the Recent tab
* Delete shared trips older than 24 hours
* Make it possible to render the ArrivalAndDeparture controller from a Trip Deep Link object
* Start overhauling OBANavigationTarget
* Wire up navigation target stuff for OBARecentStopsViewController
* Shoehorn in a way to launch differently in response to deep links.
* Unify navigation target, deep link, and 3D Touch quick action navigation systems

Miscellaneous:

* Move -escapePathVariable into OBAURLHelpers
* Fix credits - turns out that I didn't end up adding the credits.html file back to the project when I did the big project reorg last month. Oops.

Cell Swiping:

* Add third party library YMSwipeTableViewCell to allow us to display icons on swiped cells (only on the stop view controller for now!)
* Improved swipe menu for departure rows on the stops controller

Icons:

* New bookmark icons
* Massively improve tab bar icons
* Update launch storyboard - remove text and ensure that the icons are the correct, new ones
* Add full sized icons from Noun Project to the Resources folder

----

Walkthrough:

simulator screen shot nov 4 2016 1 04 45 am

simulator screen shot nov 4 2016 1 05 05 am

simulator screen shot nov 4 2016 1 05 16 am


barbeau:
The URL format is the one shown in the text message screenshot? I should be able to make this work on Android as well.

aaronbrethorst:
@barbeau here's the URL format I'm using:

```
https://www.onebusaway.co/regions/<region_id>/stops/<stop_id>/arrivals
```

with query params:

- `trip_id`
- `service_date`
- `stop_sequence`

barbeau:
Thanks!  What's the significance of the `/arrivals` portion of the URL?

aaronbrethorst:
@barbeau 

> Thanks! What's the significance of the /arrivals portion of the URL?

They map to different actions. Try opening https://www.onebusaway.co/regions/1/stops/1_11420/arrivals?trip_id=1_30938281&service_date=1478242800000&stop_sequence=4 vs https://www.onebusaway.co/regions/1/stops/1_11420 on your desktop.

barbeau:
Wow - I'm horrified to admit I didn't even know that existed.  Were the `www.onebusaway.co` URLs previously used in OBA iOS for something else?  I had never seen them before this PR.

aaronbrethorst:
No reason to be horrified, I registered the domain last week :)

```
Domain Registration Date: Fri Oct 28 03:58:20 GMT 2016
```

I needed something I could throw a simple Rails app on for this sort of purpose. Happy to transition it to onebusaway.org at a later date, but I figured that would be a lengthier process than slapping up something quick.

barbeau:
Ok, that makes me feel better :).  Any objection to using this for other OBA purposes?  We were thinking about a centralized URL scheme related to the Microsoft Research Embedded Social features, and it would be great to leverage what you've already created.

aaronbrethorst:
No, not at all. I was hoping we could, actually!

barbeau:
That's great!  Thanks for implementing this, I think it will be very useful.

barbeau:
After looking at this a little more closely - IMHO the current `/arrivals` endpoint seems to be a bit of a misnomer. In OBA Android,  the corresponding screen that you would be taken to is the trip details activity. 

Same with https://www.onebusaway.co/regions/1/stops/1_11420/arrivals?trip_id=1_30938281&service_date=1478242800000&stop_sequence=4, on the web it takes you to the trip.action site. 

So it seems a more natural name for this endpoint would be `/trips` - either higher up in the hierarchy after regions, or in place of `/arrivals`. 

However, I currently know nothing of how OBA.co is implemented. 

@aaronbrethorst Thoughts on this? I'm happy to stick with the current structure if it causes a lot of work, just my two cents looking at this for the first time.

aaronbrethorst:
A renaming would be trivial. Happy to do it. `trips` sounds good.

aaronbrethorst:
@barbeau I just deployed a change that changes `arrivals` to `trips`. Feel free to use as you see fit. Additionally, if you could create a new repository, then I'll push the app code up to GitHub.

barbeau:
Thanks!  Sure, I can create new repo - what do you want the name to be?

aaronbrethorst:
I'm not particularly emotionally attached to any particular name. `onebusaway-deep-links`, `onebusaway-webapp`, `onebusaway-webv2`, `obaco`, or anything else would be fine with me.

barbeau:
Done!   https://github.com/OneBusAway/onebusaway-deep-links

PR #857, opened 2016-11-06
Fix analytics

### LINKED ISSUE #849: Investigate drop-off in analytics numbers
There's a recent sharp drop-off in iOS usage on Google Analytics. I'm guessing this must be the result of a misconfiguration somewhere in the iOS codebase.

aaronbrethorst:
(and give users a way to opt out of analytics again too)

Fixes #849 - Investigate drop-off in analytics numbers

* At some point, the ‘opt out of analytics’ flag ended up always being set to true. Doh, my bad. I’ve fixed this, and renamed it so that we can avoid not not having double negatives.
* Add a ‘Settings’ view controller to consolidate all of the toggles we have to have in the app.
* Make it possible to get to Settings from the Info tab

### DIFF
--- a/OBAKit/Helpers/OBACommon.h
+++ b/OBAKit/Helpers/OBACommon.h
@@ -27,6 +27,9 @@ extern NSString * const kApplicationShortcutMap;
 extern NSString * const kApplicationShortcutRecents;
 extern NSString * const kApplicationShortcutBookmarks;
 
+// User Defaults Keys
+extern NSString * const OBAOptInToTrackingDefaultsKey;
+
 /**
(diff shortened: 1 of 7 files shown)


barbeau:
@aaronbrethorst Is this fix released to production?  iOS analytics numbers still seem low - I'm showing around 21k iOS users over last month.  Android is almost 170k.

aaronbrethorst:
This fix will be in the next release to go to the App Store, which I am targeting handing off to Apple on December 29.

barbeau:
Ok, thanks!

PR #858, opened 2016-11-07
Make links in service alerts tappable

### LINKED ISSUE #848: Allow tapping on URLs in Service Alerts
### Expected behavior

Allow me to tap on URLs within service alerts to go to that page

### Actual behavior

URLs are treated at plain text, and tapping on them doesn't do anything.  Granted, in this example below the text doesn't include full URLs with `http://`, but just `www.gohart.org`.  On Android this is auto-linked by the TextView UI component and is tappable.

Also related - there is a separate `url` field in the API responses for service alerts (see https://github.com/OneBusAway/onebusaway-android/issues/710) - I don't know if this is supported in OBA iOS.  The main catch here is that the OBA admin website doesn't currently allow publishing a URL to that field, so HART can't populate it in service alerts.  However, `url`will still appear in other regions that are publishing service alerts with other tools - for example, see San Diego.

1. Go to Tampa region
1. Tap on any stop
1. Tap on "View Service Alerts"
1. Tap on "Flamingo Fares" service alert, and tap on "www.gohart.org"

image

### App Version/Hardware

iPhone 7 w/ iOS 10.0.2

aaronbrethorst:
Fixes #848 - Allow tapping on URLs in Service Alerts

* Create a service alert details controller
* Display the body of the service alert in the new details controller in a text view
* Use the built-in support for 'data detectors' to render links
* Intercept link taps and display them in an in-app browser instead of launching Safari
* Revise the icons used for service alerts to make them look a bit more modern


barbeau:
👍

PR #865, opened 2016-11-10
Stop showing already-departed buses

### LINKED ISSUE #864: Improve explanation of negative minutes
- [ ] Design
- [ ] Implementation
-----

img_1371

aaronbrethorst:
Fixes #864

### DIFF
--- a/OneBusAway/ui/stops/OBAStopViewController.m
+++ b/OneBusAway/ui/stops/OBAStopViewController.m
@@ -47,7 +47,7 @@ static CGFloat const kTableHeaderHeight = 150.f;
     if (self) {
         _reloadLock = [[NSLock alloc] init];
         _stopID = [stopID copy];
-        _minutesBefore = 5;
+        _minutesBefore = 0;
         _minutesAfter = 35;
     }
     return self;


barbeau:
;)

https://groups.google.com/d/msg/onebusaway-developers/pVxORUgLhe0/_e01oXDEYSUJ

aaronbrethorst:
That's pretty funny. I'd forgotten about that :) I still think it has merit, but if the majority of users find it confusing, I'm not going to argue with them. I see enable/disable checkboxes as being a path of last resort. I'll do it if necessary, but don't want to go right for that.

barbeau:
Yeah, I get it, I had the same internal dialog on Android.  My final conclusion was I didn't want to yank something from the UI that had been there from the beginning because of a (relatively) small number of vocal users complaining about not understanding it.  My guess was that far more users found it valuable than confusing, but we just hear about the users that find it confusing (many of which may be new to the app and adjusting to the entire UI).  But, that's just a guess - it does come down to a estimate of what the majority of users think.  Ultimately it's your call on iOS.

aaronbrethorst:
If I get complaints in the opposite direction I will certainly put it back. I just want to minimize the amount of reading that users have to engage in in order to get the information they want.

PR #866, opened 2016-11-10
Remove seemingly duplicated data on stops by filtering out arrivals

### LINKED ISSUE #617: Hard to interpret arrivals and departures at some stops
Bug originally from https://twitter.com/philaros/status/719679000918908928
### Expected behavior

Arrivals and departures are never double-listed at stops.
### Actual behavior

Occasionally stops do display both arrivals and departures.

Open stop `41255` in the Puget Sound region, or stop 1420 in Puget Sound.

philaros_2016-apr-11
### App Version/Hardware

2.5.1

aaronbrethorst:
Fixes #617

Instead of showing and delineating arrivals and departures, I'm trying the tack of simply filtering out arrivals altogether. Let's see if it works out alright.


barbeau:
Just FYI, you'll remove incoming vehicles to a transit center that are part of the same block.  So if I'm waiting for Route 5 Southbound at a transit center, knowing when Route 5 Northbound arrives is useful, especially if I can board the vehicle if it arrives prior to the scheduled departure time.  Some users might not like this, if they are used to it.  I believe OBA should correctly pass the schedule deviation down the block (through the transit center and between trips) so the user should still be able to see the time their Route 5 Southbound should depart.

PR #878, opened 2016-11-21
WIP Update Translation

themonki:
Update Translation

* Adding files for base language (en)
* Update spanish translations
* Clean files that do not contain translation and do not need it

**I have added the files `OneBusAway/en.lproj/Localizable.strings` because when exporting to xliff files the base language, this was taking values in Spanish.**

The ideal would be to change the keys as it is doing in [OBA Android](https://github.com/OneBusAway/onebusaway-android/blob/master/onebusaway-android/src/main/res/values/strings.xml) for `OneBusAway/en.lproj/Localizable.strings`. This way we can easily update the languages by comparing the files.

For example, we have these keys to identify a string:

``
Some information about how you use this app is sent to Google Analytics in non-personally identifiable form to help us better understand how to improve the app. To learn more, please read our Privacy Policy.
``

The key can be:

``
msg_information_google_analytics
``

So on [code](https://github.com/OneBusAway/onebusaway-iphone/blob/develop/OneBusAway/ui/info/OBASettingsViewController.m#L46) we can use:

```java
    analyticsSection.footerView = [OBAUIBuilder footerViewWithText:NSLocalizedString(@"msg_information_google_analytics", @"Analytics explanation on the Settings view controller.") maximumWidth:CGRectGetWidth(self.tableView.frame)];

```

I can update this keys, and later you can review on this PR.

### DIFF
--- a/OBAKit/Helpers/OBADateHelpers.m
+++ b/OBAKit/Helpers/OBADateHelpers.m
@@ -50,7 +50,7 @@
 + (NSString*)formatMinutesUntilDate:(NSDate*)date {
     double minutesFrom = [date minutesFrom:[NSDate date]];
     if (fabs(minutesFrom) < 1.0) {
-        return NSLocalizedString(@"NOW", @"");
+        return NSLocalizedString(@"msg_now", @"");
     }
     else {
         return [NSString stringWithFormat:@"%.0fm", minutesFrom];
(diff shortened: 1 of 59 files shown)


aaronbrethorst:
If you rebase on top of the latest bits in develop, you should be able to get this to pass.

themonki:
I rebased and force push, but still fail.

themonki:
I push a commit where change the text and message for 'keys'. The keys are generic I think it is necessary to define a format to easily identify where it is used and the content of the message.

These keys can be easy change using [commands on terminal](http://stackoverflow.com/questions/11392478/how-to-replace-a-string-in-multiple-files-in-linux-command-line#answer-21335519) for example.

I check a bit as it has been left visually, except error messages.

aaronbrethorst:
Merged after squash with #879

PR #879, opened 2016-11-26
Update Translation

aaronbrethorst:
* Adding files for base language (en).
* Update spanish translations.
* Clean files that do not contain translation and do not need it.

Change keys for translations

* changes all text and messages on code by keys for more easy way to update translations


aaronbrethorst:
@themonki - I have taken the liberty of squashing all of your commits into a single one. I think this looks good, and I really appreciate that you have taken the time to rename strings with more appropriate keys. Going forward, I'd love to adopt more of the practices outlined in this article: https://www.objc.io/issues/9-strings/string-localization/

especially the recommendations like dotted namespaces and different tables. thanks again for all of your help!

PR #882, opened 2016-12-13
Add the words "depart" and "arrive" to table cells on stops

### LINKED ISSUE #617: Hard to interpret arrivals and departures at some stops

aaronbrethorst:
Fixes #617 - Hard to interpret arrivals and departures at some stops

Rewrite the code that generates the strings used to tell the user when their bus is arriving or departing. Also, overhaul the handling of localization within OBAKit to make sure that we actually use the lproj files that are located within the OBAKit framework.

Localization:

* Custom localization macro for OBAKit.
* Migrate all of OBAKit over to the OBALocalized macro.

Stops Controller

* Better text for informing the user if a bus is arriving or departing.
* Better sorting for stops with arrivals and departures. For a bus that arrives and then departs from a stop, make sure that its arrival is always shown first.

Models:

* Plumb arrival/departure state into OBAUpcomingDeparture.
* Hide a bunch of unused properties on OBAArrivalAndDepartureV2.h within its .m file.
* Easier access to the best arrival *or* departure date for a bus with `-[OBAArrivalAndDepartureV2 bestArrivalDepartureDate]`.
* 

Misc:

* Remove an unused method from OBACommon.
* Move a few files into OBAKit that will be nice to have in there in the future when the Today screen extension gets built.
* Add back button image to the xcasset catalog. It's not relevant to this work, but I wanted it off my computer.


aaronbrethorst:
@themonki - I made some changes to the strings used to render stop text. Please see https://github.com/OneBusAway/onebusaway-iphone/pull/882/files#diff-9614f0dff5485e29e2ce57e7a1ca7076R97. I'd love your help in translating these to Spanish. Thank you!

Issue #854, opened 2016-11-03
add a toggle for disabling review requests 


aaronbrethorst:
Add a toggle under the Info tab (maybe underneath an "Advanced" page or something) entitled "Don't nag me to review OneBusAway". It is off by default. If it is set to on, then `apptentiveTimer` on OBAStopViewController never gets created.

aaronbrethorst:
I have received a total of four complaints about this out of nearly 70,000 ratings prompts shown. At least two of the complaints were from iOS engineers. I was hoping one of them might fix this, especially since the new `OBASettingsViewController` should make this nearly trivial to implement. Punting from this release.

aaronbrethorst:
Fixed in #921

PR #921, opened 2017-01-01
Added a toggle for disabling review requests

ualch9:
#854 - add a toggle for disabling review requests

### DIFF
--- a/OBAKit/Helpers/OBACommon.m
+++ b/OBAKit/Helpers/OBACommon.m
@@ -25,6 +25,7 @@ NSString * const kApplicationShortcutRecents = @"org.onebusaway.iphone.shortcut.
 NSString * const kApplicationShortcutBookmarks = @"org.onebusaway.iphone.shortcut.bookmarks";
 
 NSString * const OBAOptInToTrackingDefaultsKey = @"OBAOptInToTrackingDefaultsKey";
+NSString * const OBAAllowReviewPromptsDefaultsKey = @"OBAAllowReviewPromptsDefaultsKey";
 
 const NSInteger kOBAErrorDuplicateEntity = 1000;
 const NSInteger kOBAErrorMissingFieldInData = 1001;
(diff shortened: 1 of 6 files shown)


OneBusAway/ui/info/OBASettingsViewController.m
```diff
@@ -42,6 +42,13 @@ - (void)loadData {
         [[NSUserDefaults standardUserDefaults] setBool:!analyticsValue forKey:OBAOptInToTrackingDefaultsKey];
     } switchValue:analyticsValue];
     [analyticsSection addRow:switchRow];
+    
+    // Disable review requests - issue #854
+    BOOL ratePromptValue = [[NSUserDefaults standardUserDefaults] boolForKey:OBADisallowReviewPromptsDefaultsKey];
+    OBASwitchRow *ratePromptRow = [[OBASwitchRow alloc] initWithTitle:NSLocalizedString(@"msg_disallow_review_prompt", @"A switch option's text for enabling and disabling App Store review prompts") action:^{
```
aaronbrethorst: So here you have a switch that when set to a positive value produces a negative result, right? On = Don't do something. Off = Do something. I find these negated settings hard to follow, and find this much easier to understand if the setting was, instead, for "Prompt me for app feedback", and was set to `YES` by default.

You can register default values in OBAApplication.m. Check out the `-registerAppDefaults` method.

OBAKit/Helpers/OBACommon.m
```diff
@@ -25,6 +25,7 @@
 NSString * const kApplicationShortcutBookmarks = @"org.onebusaway.iphone.shortcut.bookmarks";
 
 NSString * const OBAOptInToTrackingDefaultsKey = @"OBAOptInToTrackingDefaultsKey";
+NSString * const OBADisallowReviewPromptsDefaultsKey = @"OBAAllowReviewPromptsDefaultsKey";
```
aaronbrethorst: These should both be `OBAAllowReviewPromptsDefaultsKey` per my other comment.

aaronbrethorst:
looks great, thank you @ualch9!

PR #922, opened 2017-01-03
[Fixes #838] Turn on Bitcode for OBA

### LINKED ISSUE #838: Get Bitcode working
It may be as simple as ticking a box in OBAKit and OneBusAway, or it may involve some mucking around with Carthage projects. Either way, we should make sure that Bitcode is enabled.

sethfri:
We were overriding the default ENABLE_BITCODE value with "No". This eliminates the override and returns to the default value.

This SHOULD enable Bitcode for the app, unless potentially one of our Carthage dependencies doesn't have it turned on. @aaronbrethorst agreed to test and see if this works during archival/submission.


aaronbrethorst:
This looks perfect! I'll merge as soon as you can sign the CLA. Thanks again!

aaronbrethorst:
👍 hooray, thanks Seth!

PR #926, opened 2017-01-05
Fix Analytics (again, but for real this time)

### LINKED ISSUE #849: Investigate drop-off in analytics numbers

aaronbrethorst:
Fixes #849

The default Analytics tracker was not being created. This is now fixed.

* Move app-specific keys (Analytics and Apptentive) into OBAApplication. This will be useful down the road if/when transit agencies want to rebrand OBA. Having these pieces of app-specific information consolidated helps us achieve this goal.
* Remove the ifdef that would disable analytics in development. The number of developers is so minuscule compared to the number of users that this will not affect our stats at all. Plus, it reduces the potential risk of analytics failing again in the future.
* Remove an unsightly batch of `_Nonnull` annotations by setting `NS_ASSUME_NONNULL_BEGIN` sooner in OBAAnalytics.h.

### DIFF
--- a/OBAKit/Info.plist
+++ b/OBAKit/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundleShortVersionString</key>
 	<string>17.1.1</string>
 	<key>CFBundleVersion</key>
-	<string>20170104.09</string>
+	<string>20170105.12</string>
 	<key>NSPrincipalClass</key>
 	<string></string>
 </dict>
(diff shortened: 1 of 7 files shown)


barbeau:
Thanks! Is this going out in the release you just put together? I'd like to quote numbers at TRB next week, but that requires this fix.

PR #929, opened 2017-01-06
Updated transport glyphs

ualch9:
Updated glyphs to its Android counterpart, I updated these icons because they appear too complex and confusing when it is small and because of inconsistency.


aaronbrethorst:
Awesome, looks great

barbeau:
Yay material design! ;)

Issue #928, opened 2017-01-06
Make it possible to copy just the URL for trip sharing


aaronbrethorst:
Let the user copy only the URL to the clipboard.

PR #930, opened 2017-01-07
Choosing "Copy" will copy only the URL

ualch9:
#928


ualch9:
Fixed per your feedback `55e2a65`

aaronbrethorst:
nice, thank you! 👍

PR #948, opened 2017-01-18
Rely on standard UIButton behavior to drive FAB

aaronbrethorst:
Remove a bunch of code that duplicates some lesser known features of UIControl and UIButton


ualch9:
:+1:

PR #972, opened 2017-02-07
Fixes handoff not properly working for deep link URLs

### LINKED ISSUE #966: Handoff stops does not work
Just dumping some stuff in this thread for now, will further investigate

Region on both devices are both set to Puget Sound

ualch9:
Fixes #966


aaronbrethorst:
👍

PR #997, opened 2017-03-11
Improve Map and Search User Experience

aaronbrethorst:
The search user experience has been in dire need of an overhaul for years. This commit overhauls the search experience to make it easier and faster to find a desired route or stop number or a particular address. It also makes it possible to search your bookmarks and recent stops for an even quicker, all-in-one experience.

The map Ux, likewise, hasn't been changed in any meaningful way in years. This commit follows the design set forth by every other mapping app and moves critical buttons to the bottom right corner of the map view. This also allows for more space to devote to the search box.

Lastly, this commit greatly simplifies the map view controller's code, which will make it far easier to accomplish other critical tasks on the map going forward.

Other stuff:

* Rename OBASearchController to OBAMapDataLoader, which better describes its purpose
* Rename OBASearchResultsMapViewController to OBAMapViewController because the extra two words weren’t giving us any value
* Remove the Map Controller xib file
* Remove landscape orientation support from the app on iPhone. It’s pointless and frustrating.
* Remove Pulley and drawer ui concept. Drawer UI is cool, but not being kept up with changes across rest of app.
* Delete OBASearch; move its methods into a category on OBANavigationTarget

------

img_2209

img_2210


barbeau:
:+1:  - nice!

ualch9:
:+1:

themonki:
Cool!!!, tell me when I can do the translation.

aaronbrethorst:
@themonki thank you, will do!

PR #1022, opened 2017-04-01
Regional alerts

### LINKED ISSUE #881: Add a stop-gap service alert solution to OBA for Puget Sound
The Puget Sound OBA deployment currently does not include service alerts. Sound Transit does have an RSS feed that includes a more 'free-form' list of service alerts. Until ST is able to pipe this data into OBA, it would be nice to have a way to give riders a heads up about possible issues, even if we can't directly map those alerts to specific stops or routes.

Giving users the ability to learn whether their route has been canceled, redirected because of snow, or is on reduced service because of a Sounders parade is an extremely valuable service to offer. 

The XML document is almost 2MB in size, and we can't have people downloading that tens of thousands of times a day. To that end, the file will have to be downloaded by OBA.co at regular intervals, parsed there, and stored for (partial) re-transmission to OBA clients. Maybe as a JSON file on the other side, because I'd much rather write JSON parsing code on iOS than XML parsing code.

- [x] OBA.co RSS feed reader/model/persistence (http://www.soundtransit.org/schedules/alerts.xml)
- [x] OBA.co API for global service alerts corresponding to a region. (right now only region `1` will have them)
- [ ] Downloader, modeling, and persistence for global service alerts. (I'd love to reuse the existing service alert model.)
- [ ] Design for global alert UI
- [ ] Implement global alert UI

aaronbrethorst:
Fixes #881 

* Add new model service methods to make API calls into the regional alerts service
* OBARegionalAlert model, service object, and associated plumbing
* Add an Alerts table item to the Info controller and a view controller to render regional alerts
* Add OBAMessageCell to alerts table view look sort of like the Mail app

Misc:

* Change how notifications are registered in App Delegate
* Modify Swift optimization level for Debug on OBAKit to fix debugging issues


bbodenmiller:
I don't have an iOS device anymore but I'd love to see what the UI looks like here. Also did you see KCM supports IFTTT now? https://ifttt.com/king_county_metro

bbodenmiller:
Also does Android now have the stop-gap solution for Puget Sound also? I've been out of the region recently.

barbeau:
Android only supports alerts if they are published directly via the OBA server.

PR #1032, opened 2017-04-03
Update Spanish Translations

themonki:
Update Spanish Translations

* Based on commit c47898b - Fixes format string appearing in vehicle map popover
* Based on commit ecac56f - Adds ‘secret’ debug mode to the app
* Based on commit 6ad30a7 - Only show experimental regions on the Region List controller when debug mode is enabled
* Based on commit 2cceef0 - Adds Regional Alerts
* Based on commit 8d852f8 - Fixes display bug for Nearby Stops without specified direction
* Based on commit 18f790a - Allows sorting bookmarks by proximity


aaronbrethorst:
awesome!

PR #1060, opened 2017-04-18
Update spanish translations

themonki:
Update spanish translations

* Based on commit f03317d - Adds a location services onboarding controller
* Based on commit 279e0f7 - Adds user defaults debugging tools to the Info tab
* Based on commit c2d59e4 - Fixes stop number lookup issue
* Based on commit c6f5e4a - Ensure that push notification tokens are always refreshed before sending them to the server


aaronbrethorst:
awesome, thank you!

PR #1098, opened 2017-04-27
Displays a "close" button in addition to existing behavior for dismissing an alert

### LINKED ISSUE #1094: AlertPresenter alerts lack a 'dismiss' affordance
I received feedback from a user who mentioned that waiting for an `AlertPresenter`-based alert to time out was tedious, and I realized for the first time that the `AlertPresenter` lacks any 'Dismiss' affordances! You can tap anywhere on the screen to dismiss one of these, but there's nothing about them that suggests to the user that this is possible, which is a huge design issue and my bad :-\

A good solution for this will include some sort of 'Dismiss' button

- [x] needs design
- [ ] needs implementation 

attachment 3

ualch9:
fixes #1094


aaronbrethorst:
awesome, thank you 👍

aaronbrethorst:
awesome

Issue #1073, opened 2017-04-23
need a better icon for 'nearby stops'


aaronbrethorst:
The current icon is ugly and makes no sense.

![ick](https://cloud.githubusercontent.com/assets/2254/25309945/2feb573a-278e-11e7-836b-faea82de7e61.jpg)

PR #1116, opened 2017-05-04
Update nearby icon.

cathy810218:
Fixes https://github.com/OneBusAway/onebusaway-iphone/issues/1073 - need a better icon for 'nearby stops'

* Add 3 different icons for nearby button but currently set to use the thrid one. Please let me know which one you prefer and I will change it.


ualch9:
I think these icons are confusing, I showed these to several other people who are not sure what all 3 icons meant. Just a thought.

aaronbrethorst:
thank you for your feedback, Alan. @cathy810218 - I'll follow up on this and the other PRs after work today.

cathy810218:
Okay! Yea, I had my sister draw them. And they did not make sense to me at
first. I thought it was just me. but I guess we will try to think of other
designs. Thank you for the feedback :)

ualch9:
@cathy810218 I had thought about using POI icons such as the image attached, but never got the time to mock them up. Please let me know what you think.

barbeau:
Do we have an idea of how often users switch between the nearby stops map vs. list view?  I haven't looked at the analytics data for this on iOS. We have the same design issue on Android although we haven't implemented the nearby stops list yet.

cathy810218:
How about this one?

simulator screen shot may 6 2017 9 21 16 pm

dotTechnopath:
for reference, this is what google uses for their "nearby places" icon (seen on google maps)
image

cathy810218:
How about this one?
nearby

cathy810218:
simulator screen shot may 6 2017 9 50 17 pm

barbeau:
Assuming the functionality here is flipping between the view of nearby stops as a list or a map, I'm not sure the icon itself is really the problem.  I think it's just not clear that this is a toggle control specific to viewing nearby stops two different ways. 

How about renaming the bottom tab to "Nearby" instead of "Map" and change that icon to a placemark icon?  I think this helps give the overall screen a better context - in other words, this isn't just a map view of something. Then the question is how to show a good toggle control between map and list.

dotTechnopath:
I agree with @barbeau , however im not really sure what the solution would be.
one idea is to integrate the nearby stops with the search bar (this is similar to apple maps and also google maps to an extent). So when a user taps on the search bar, instead of having a blank canvas, there will be nearby stops listed under the search bar. 
another idea is to combine the nearby and the recent stops tab, and having a toggle within that tab (similar to the bookmarks tab)

also, I disagree with the wording of "nearby" on the maps tab. the wording i think could potentially be unclear to the users. Users might take it literally and question if the "nearby" tab will _only_ show nearby stops

aaronbrethorst:
Closing in favor of #1251

Issue #1091, opened 2017-04-25
Fix z order for annotations and callouts on the map


aaronbrethorst:
### Expected behavior

Annotations and callouts on the map should have a reasonable z order.

- [ ] Define proper behavior
- [ ] Implement it

### Actual behavior

it's not very reasonable at present.

### Screenshot of the issue (if applicable)

![attachment 2](https://cloud.githubusercontent.com/assets/2254/25411765/6b2546d4-29d3-11e7-83b5-86c49674c147.jpg)

PR #1117, opened 2017-05-04
Push back user location blue dot to prevent blocking call out view.

cathy810218:
Fixes https://github.com/OneBusAway/onebusaway-iphone/issues/1091 - Fix z order for annotations and callouts on the map

* Define behavior: The blue dot should always be below the call out views,
so it won't block the call out views when user location gets updated.
User will never lose track of their own location because we have a button that
takes them back to where they are.

* In the OBAMapViewController regionDidChangeAnimated, instead of
bringSubviewToFront, I use sendSubviewToBack to push back the blue dot.

### DIFF
--- a/OneBusAway/ui/map/OBAMapViewController.m
+++ b/OneBusAway/ui/map/OBAMapViewController.m
@@ -365,7 +365,7 @@ static const double kStopsInRegionRefreshDelayOnDrag = 0.1;
 - (void)mapView:(MKMapView *)mapView regionDidChangeAnimated:(BOOL)animated {
     if (mapView.userLocation) {
         UIView *annotationView = [mapView viewForAnnotation:mapView.userLocation];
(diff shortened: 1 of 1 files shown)


OneBusAway/ui/map/OBAMapViewController.m
```diff
@@ -568,7 +568,7 @@ - (void)displayStopControllerForStopID:(NSString*)stopID {
 
 - (void)refreshCurrentLocation {
     CLLocation *location = self.locationManager.currentLocation;
-
+
```
aaronbrethorst: Mind removing this whitespace and squashing the second commit?

aaronbrethorst:
Squashing: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

PR #1118, opened 2017-05-05
Push back user location blue dot to prevent blocking call out view.

cathy810218:
Fixes https://github.com/OneBusAway/onebusaway-iphone/issues/1091 - Fix z order for annotations and callouts on the map

* Define behavior: The blue dot should always be below the call out views,
so it won't block the call out views when user location gets updated.
User will never lose track of their own location because we have a button that
takes them back to where they are.

* In the OBAMapViewController regionDidChangeAnimated, instead of
bringSubviewToFront, I use sendSubviewToBack to push back the blue dot.


cathy810218:
@aaronbrethorst Thank you! I tried to squash the commits, but then it still showed up as I removed the space. I just wanted to discard that extra space change, so instead I did `reset HEAD` and force push to my branch.

aaronbrethorst:
Why'd you close this?

cathy810218:
I want Travis to rebuild it.

aaronbrethorst:
I manually kicked off a new travis build from inside of travis. dunno if you have access to the button. feel free to poke me to restart travis when needed.

Issue #1017, opened 2017-03-31
Map type in header on the stop controller should match map's map type


aaronbrethorst:
### Expected behavior

If I use the `hybrid` map type on the Map tab, that choice should be reflected in the headers on stop controllers.

### Actual behavior

It always uses the `standard` type instead.

### Steps to reproduce the behavior

1. Switch the map tab to the hybrid style
2. navigate to a stop

### App Version/Hardware

17.2

PR #1119, opened 2017-05-06
Display the correct map image in Stop Controller

cathy810218:
Fixes https://github.com/OneBusAway/onebusaway-iphone/issues/1017 - Map type in header on the stop controller should match map's map type

* Get the stored mapType from NSUserDefault to show the correct header image in Stop Controller.


aaronbrethorst:
👍

Issue #1120, opened 2017-05-06
Alarm button title doesn't get updated after it gets removed.


cathy810218:
### Expected behavior

After setting alarm, on the same view controller, the alert sheet should have "Remove Alarm".
After removing alarm, on the same view controller, the alert sheet should have "Set Alarm".

### Actual behavior

After adding the alarm, on the same view controller, select the option button (green button with three white dots) again, the alert sheet still has "Set Alarm" but it behaves as remove alarm.

After removing, and on the same view controller, click the option button, the alert sheet still has "Remove Alarm" button. But when select it, it behaves as set alarm.

### Steps to reproduce the behavior

1. Select a stop, and tap on the option button, and choose "Set Alarm".

2. And tap on the same button again, it still shows "Set Alarm" but when you select it, it will remove your alarm.

### App Version/Hardware
17.4.0

PR #1122, opened 2017-05-06
Manually trigger reloadData to update alarm button title in alert sheet

cathy810218:
Fixes https://github.com/OneBusAway/onebusaway-iphone/issues/1120 - Alarm button title doesn't get updated after it gets removed

* After setting or removing alarm, call `updateAlarmStateForRowAtIndexPath:indexPath` to update the alarm state. (i.e if `alarmExists` is `NO`, set it to`YES`; if `alarmExists` is `YES`, then set it to `NO`)

* Update `indexPathForModel:model`. When comparing two `model` objects, instead of using `isEqual`, use `==` to compare their pointers (deep comparison) to get the correct `indexPath`.

Steps for reloading the row with given model:
1. When user adds/removes an alarm, we only know the model, so we will call `indexPathForModel:(id)model` to get the `indexPath`
2. With the correct `indexPath` we can get the row of type `OBADepartureRow` calling `rowAtIndexPath:`
3. Having the selected row, we need to update its current alarm state. `row.alarmExists = !row.alarmExists;`
4. Call `reloadRowsAtIndexPaths: withRowAnimation` to reload just one row.

### DIFF
--- a/OneBusAway/ui/static_table/OBAStaticTableViewController.m
+++ b/OneBusAway/ui/static_table/OBAStaticTableViewController.m
@@ -148,8 +148,7 @@
         OBATableSection *section = self.sections[i];
         for (NSInteger j=0; j<section.rows.count; j++) {
             OBABaseRow *candidate = section.rows[j];
-
-            if ([candidate.model isEqual:model]) {
+            if (candidate.model == model) { // Need to compare the pointers to get the correct indexPath
                 return [NSIndexPath indexPathForRow:j inSection:i];
             }
         }
--- a/OneBusAway/ui/stops/OBAStopViewController.m
+++ b/OneBusAway/ui/stops/OBAStopViewController.m
@@ -363,6 +363,7 @@ static NSInteger kStopsSectionTag = 101;
     NSURLRequest *request = [self.modelService.obaJsonDataSource requestWithURL:alarm.alarmURL HTTPMethod:@"DELETE"];
     [self.modelService.obaJsonDataSource performRequest:request completionBlock:^(id responseData, NSUInteger responseCode, NSError *error) {
         [self.modelDAO removeAlarmWithKey:dep.alarmKey];
+        [self reloadDataAnimated:NO];
     }];
 }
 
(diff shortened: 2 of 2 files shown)


aaronbrethorst:
Reloading all of the data in the controller to change the text of the action sheet feels kind of heavyweight to me. I'd prefer a solution where only one cell gets reloaded. Ping me on Slack if you want to talk though how that might work.

OneBusAway/ui/static_table/OBAStaticTableViewController.m
```diff
@@ -148,8 +148,7 @@ - (NSIndexPath*)indexPathForModel:(id)model {
         OBATableSection *section = self.sections[i];
         for (NSInteger j=0; j<section.rows.count; j++) {
             OBABaseRow *candidate = section.rows[j];
-
-            if ([candidate.model isEqual:model]) {
+            if (candidate.model == model) { // Need to compare the pointers to get the correct indexPath
```
aaronbrethorst: Can you elaborate on why `-isEqual:` wasn't working?

OneBusAway/ui/stops/OBAStopViewController.m
```diff
@@ -363,9 +363,19 @@ - (void)deleteAlarmForArrivalAndDeparture:(OBAArrivalAndDepartureV2*)dep {
...
     [self.modelService.obaJsonDataSource performRequest:request completionBlock:^(id responseData, NSUInteger responseCode, NSError *error) {
         [self.modelDAO removeAlarmWithKey:dep.alarmKey];
+        // Replace with a new row and update it.
+        [self updateAlarmStateForRowAtIndexPath:[self indexPathForModel:dep]];
     }];
 }
 
+- (void)updateAlarmStateForRowAtIndexPath:(NSIndexPath *)indexPath {
+    OBADepartureRow *row = (OBADepartureRow *)[self rowAtIndexPath:indexPath];
+    OBADepartureRow *newRow = row;
```
aaronbrethorst: What's this doing?

cathy810218: To not reload the entire table view, I decided to reload just a single table cell.
And I create a new row that is the same as the old one but with the new updated alarm state.
Then replace the old row and reload it.

cathy810218: If you meant where I assigned `newRow = row`, it was because I thought I needed to replace the old one with a new row. 
However, I just updated the code so now it will only grab the current row which alarm is being added/removed and update it's alarm state. Then reload this row.

aaronbrethorst:
👏 👏 👏

Issue #1121, opened 2017-05-06
Setting alarm for a past/arriving now bus causes hard crash


cathy810218:
### Expected behavior

Should not show "Set Alarm" option if the bus is arriving now or it's already gone.

### Actual behavior

If the bus is gone or arriving now, when selecting "Set Alarm", it doesn't show any thing in the "Alert Me" menu. But clicking "OK" button causes crash.

### Steps to reproduce the behavior

1. Select a stop, and tap on the option button for a "Arriving Now" or a past bus.

2. Choose "Set Alarm" and hit "OK".

### App Version/Hardware

17.4.0

PR #1123, opened 2017-05-06
Disable the option "Set Alarm" when the bus is arriving or departed.

cathy810218:
Fixes https://github.com/OneBusAway/onebusaway-iphone/issues/1121 - Setting alarm for a past/arriving now bus causes hard crash

* Create a Boolean to check if the bus is arriving/departed (including Arriving now) by setting `minutesUntilBestDeparture > 0`.

* If `hasArrived` is `YES` for the current `row`, then disable `Set Alarm` option.


aaronbrethorst:
Hmm, good catch. This definitely needs to get fixed. I like the concept you're presenting here, but since you end up negating the value of `hasLeft`, I'd prefer seeing it expressed as something more like  `hasArrived`, which is YES if `minutesUntilBestDeparture > 0`. That way, the `if` statement doesn't need a negation, which makes for clearer reading.

aaronbrethorst:
perfect :)

PR #1147, opened 2017-05-14
Add Mandrain localizable.string

cathy810218:
Add Traditional Chinese.

* We use `Localizable.strings` under `OBAKit` for `msg_scheduled_explanatory`, so I added them to `org.onebusaway.iphone`


aaronbrethorst:
Awe. Some. thank you!

Issue #1124, opened 2017-05-07
Bookmark tab's table cell actions disappear when table reloads


aaronbrethorst:
The bookmark controller's 'swipe to delete' context action disappears automatically when the table view reloads after its data updates. Either there should be a way to detect that a cell has its action exposed, thereby temporarily disabling the reload feature, or the contextual action should be removed.

PR #1152, opened 2017-05-27
Pause reloading bookmark cell when in editing mode

cathy810218:
Fixes https://github.com/OneBusAway/onebusaway-iphone/issues/1124 - Bookmark tab's table cell actions disappear when table reloads

* When user is in editing mode, set `self.editing = YES` to pause reloading data.

* When user `didEndEditing`, set `self.editing = NO` to resume reloading data.


aaronbrethorst:
Looks great.

PR #1153, opened 2017-06-01
Removes the Apptentive SDK

aaronbrethorst:
I cannot keep committing to handling 100% of app feedback + a staggering number of complaints about transit service issues. To that end, I’m removing Apptentive SDK integration. It was fun while it lasted. I’ll resurrect the ‘email the mailing list’ thing and add iOS 10.3’s ratings prompt support soon.


bbodenmiller:
☹️

Issue #935, opened 2017-01-13
The walk indicator can be wrong when you're standing at the bus stop


aaronbrethorst:
From a user:

> I'm at the stop but it says I can't make the soonest bus about to arrive

![attachment 1](https://cloud.githubusercontent.com/assets/2254/21914972/868751c6-d8eb-11e6-8683-6e0be3bc1f5e.jpg)

PR #1158, opened 2017-06-04
Fix walk indicator when user is within negligible distance to stop.

GalOshri:
If user is already at bus stop (or within negligible walking time to
it), walk indicator should show that they can reach a bus if it hasn’t
departed yet.
Fix for issue #935.


aaronbrethorst:
Looks good, thanks.

PR #1189, opened 2017-10-02
Update Spanish Translations

themonki:
* Based on commit fa1230d - Adds properly formatted accessibility labels for time until next departures on the bookmarks tab
* Based on commit 40b590 - Switches back to requesting app feedback to be sent to iphone-app mailing list
* Based on commit df529f3 - Adds option to disable the compass heading indicator on the map
* Based on commit ce0520b - Adds walking directions options to the '...' button in the Stop page
* Based on commit e1da140 - Replaces URLs in errors with SHA-1 reference IDs
* Based on commit 77f9029 - Adds feedback fallback for users not using default Mail.app

### DIFF
--- a/OBAKit/es.lproj/Localizable.strings
+++ b/OBAKit/es.lproj/Localizable.strings
@@ -4,6 +4,12 @@
 /* sd > 0 */
 "msg_space_late" = " tarde";
 
+/* One minute */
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
nice! :+1:

PR #1208, opened 2017-10-17
Add strings file to Today extension

jeffpearce:
- Add strings from main app to build for Today
- Fix crash using one of the strings


aaronbrethorst:
👍

PR #1228, opened 2017-12-18
fix deep link

jeffpearce:


aaronbrethorst:
nicely done, @jeffpearce! I love it when they turn out to be this easy!

PR #1305, opened 2018-06-05
Remove Google Dependencies

aaronbrethorst:
Removes all Google API dependencies from OBA for iOS.


aaronbrethorst:
cc @barbeau

barbeau:
Thanks @aaronbrethorst.  Is Google Maps used for anything in OBA iOS in the currently release, or is this just removing dead code?

aaronbrethorst:
@barbeau we were using a Google Maps API to geocode addresses when the user performed an address search, but Apple has had this feature built in since iOS 6 without any usage limits, so I thought it made more sense to use what Apple offered, reduce external dependencies, and avoid usage limits. Here's the relevant diff https://github.com/OneBusAway/onebusaway-iphone/pull/1305/files#diff-9b0c079617a4cdd9ac1b1211a4a1b0d8R261

barbeau:
👍

PR #1335, opened 2018-08-03
Update the Spanish Translations

themonki:
* Based on commit #e94ba6d - Embed a stripped-down stop controller in the map's callouts
* Based on commit #467fd7a - Adds Dark Sky attribution
* Based on commit #b9b183d - Make it possible to search for a vehicle by vehicle id
* Based on commit #e4f4bdc - Display stops in a Pulley drawer on the map


aaronbrethorst:
Wonderful, thank you @themonki!

PR #1366, opened 2018-09-15
Overhauls Analytics

### LINKED ISSUE #1327: Better Fabric analytics on accessibility usage
It would be great if we could track the number of folks using voiceover and high contrast settings on a per-app version basis, plus also know what the percentage mix of font sizes is.

### LINKED ISSUE #1343: Analytics for Fare Payment apps
TODO: need to sort out with @barbeau 

1. Track taps on Pay My Fare table item -> App Store
2. Track taps -> apps

anything else?

aaronbrethorst:
* Fixes #1327 - Better Fabric analytics on accessibility usage
* Fixes #1343 - Analytics for Fare Payment apps

This change adds support for Firebase Analytics to the app, which is necessary as Fabric will be fully subsumed into Firebase in the next six months or so. Additionally, it increases the number of accessibility-related data points that we are tracking—we will now know what users' default font sizes are and whether they have high contrast mode enabled.


barbeau:
Nice! I'm actually in the middle of adding Firebase Cloud Firestone on Android to collect a limited amount of data during beta testing of destination alerts. Have you seen if iOS and Android apps should share same Firebase account as far as best practices? If so we should coordinate - I don't have anything stored in my setup yet so we could switch to yours if that makes sense.

PR #1461, opened 2019-12-25
Fix search not working when user location is not within bounds of the current region

### LINKED ISSUE #1456: search not working when far from current location
Search is working OK for Puget Sound.  However, when I try to do a search for a route or stop in other regions the map says "No stops at this location", even though there are.  I tried San Diego, Tampa, and Spokane, with the same result.

The two screenshots show this for Tampa.  The first image shows the map for a randomly chosen stop (Marion St @ Zack St).  Routes 8 and 20 stop there.  Then, keeping the map at the same location, I search for route 8, and get the second image, which says "No stops at this location" even though clearly there are.

The same thing happens for searching for a stop number.  App 19.1.0, running on an iPhone 7, iOS version 13.1.3

IMG_3006

IMG_3007

### LINKED ISSUE #1458: Search doesn’t work
### Expected behavior
When searching for route, it should show the route on map
### Actual behavior
Upon search of route, it says stop is notf found
Newest version from App Store
Select your region
Zoom in to area with your desired route
Search for that route using search function and route section of search 

https://streamable.com/a0ui4

### App Version/Hardware
iOS 19.1.0

ualch9:
- Fixes #1456
- Fixes #1458


aaronbrethorst:
👍 nicely done

PR #1462, opened 2020-01-06
Add dark mode support for CreditViewerController

ApolloZhu:
Replace deprecated UIWebView with WKWebView. See https://developer.apple.com/news/?id=12232019b for deprecation notice.

|Before|After|
|--|--|
|Before|After|


aaronbrethorst:
thanks!

PR #1472, opened 2020-03-01
Resolves #1466

### LINKED ISSUE #1466: User feedback: Today View Widget not showing bookmarks
User feedback:

> Widget does not work or refresh. Just always shows this. 

today

```
App Version = 20.1.0
iOS Version = 13.3
Bookmark Count = 1
```

ualch9:
Adds the number of bookmarks in the today widget to debug info.

I'm guessing that #1466 is user error so if there are any future inquiries regarding the today widget, hopefully this data makes it easier to address.

### DIFF
--- a/OBAKit/Models/dao/OBAModelDAO.m
+++ b/OBAKit/Models/dao/OBAModelDAO.m
@@ -376,6 +376,14 @@ const CLLocationDistance kMetersInOneMile = 1609.34;
     }
 }
 
+- (NSInteger)todayBookmarkGroupCount {
+    if (self.todayBookmarkGroup) {
+        return [self.todayBookmarkGroup.bookmarks count];
+    } else {
+        return 0;
+    }
+}
+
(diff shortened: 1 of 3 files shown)


OBAKit/Models/dao/OBAModelDAO.m
```diff
@@ -376,6 +376,14 @@ - (OBABookmarkGroup*)todayBookmarkGroup {
     }
 }
 
+- (NSInteger)todayBookmarkGroupCount {
+    if (self.todayBookmarkGroup) {
+        return [self.todayBookmarkGroup.bookmarks count];
```
aaronbrethorst, line 381: Being explicit is ok, but for future reference, Objective-C treats `nil` and `0` as being essentially interchangeable—and it also treats the result of messaging `nil` as `nil`—so you could shorten this method to just:

    - (NSInteger)todayBookmarkGroupCount {
        return self.todayBookmarkGroup.bookmarks.count;
    }

aaronbrethorst (approved):
👍

PR #1475, opened 2020-04-15
Trip Status Modifier

### LINKED ISSUE #1474: Support CANCELED trips
We need to do the same as what is described here for Android:

https://github.com/OneBusAway/onebusaway-android/issues/1033

### LINKED ISSUE #1476: Memory hogging/leak
Occurs when pushing new `OBAArrivalAndDepartureViewController` from `OBAStopViewController`.

Screen Shot 2020-04-15 at 9 25 16 PM

ualch9:
Fixes #1474 
Closes #1476 

- [x] Add status property to OBATripStatusV2
- [x] UI reflects trip status
- [ ] Test with real production data

### DIFF
--- a/OBAKit/Models/unmanaged/OBAArrivalAndDepartureV2.m
+++ b/OBAKit/Models/unmanaged/OBAArrivalAndDepartureV2.m
@@ -156,6 +156,10 @@
 #pragma mark - Public methods
 
 - (OBADepartureStatus)departureStatus {
+    if (self.tripStatus.isCanceled) {
+        return OBADepartureStatusCanceled;
+    }
+
     if (!self.hasRealTimeData) {
         return OBADepartureStatusUnknown;
     }
--- a/OBAKit/Models/unmanaged/OBADepartureStatus.h
+++ b/OBAKit/Models/unmanaged/OBADepartureStatus.h
@@ -12,5 +12,6 @@ typedef NS_ENUM(NSUInteger, OBADepartureStatus) {
     OBADepartureStatusUnknown = 0,
     OBADepartureStatusEarly,
     OBADepartureStatusOnTime,
-    OBADepartureStatusDelayed
+    OBADepartureStatusDelayed,
+    OBADepartureStatusCanceled
 };
--- a/OBAKit/Models/unmanaged/OBATripStatusV2.h
+++ b/OBAKit/Models/unmanaged/OBATripStatusV2.h
@@ -21,6 +21,21 @@
 #import <OBAKit/OBATripV2.h>
 #import <OBAKit/OBAFrequencyV2.h>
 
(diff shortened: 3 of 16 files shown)


ualch9:
Simulator Screen Shot - iPhone 8 - 2020-04-14 at 21 13 08

barbeau:
@ualch9 I like the strike-through text in the arrivals list!

@ualch9 @aaronbrethorst One UI decision I need to make on Android, and I believe you do here on iOS as well - do we show vehicle markers on the map for canceled trips?

On Android, we show all vehicle markers for a given route on the map at the same time. 

image

Scheduled vehicles without real-time info show up as gray markers with "SCHEDULED" in place of the delay value.

In this context, I'm leaning towards not showing the vehicle marker at all on the map if the trip is canceled. I believe this will be easier to understand to users than showing a vehicle marker with some kind of CANCELED annotation.

Any thoughts on iOS?

ualch9:
@barbeau
I was planning:

1) Don’t let the user tap on a canceled trip in the arrivaldeparture tableview, display an alert instead informing the user that the trip is canceled.
2) If the user uses a deep-link, kick the user out of the vehicle marker map view and show a ‘trip canceled’ alert.

Also, are there any oba deployments that have implemented the CANCELED trip status modifier?

ualch9:
Screen Shot 2020-04-15 at 10 39 07 PM

ualch9:
@aaronbrethorst please review

barbeau:
>Also, are there any oba deployments that have implemented the CANCELED trip status modifier?

Not to my knowledge at this point. It sounds like MTA NYC may have this in production for their BusTime, but they aren't a production region for the native OBA apps, and mileage-may-vary when trying the native OBA apps with them (see http://regions.onebusaway.org/regions-v3.json for URL).

barbeau:
So I may be wrong on the above - Puget Sound may have trips with `CANCELED` status live now. Please post the URL and responses if you find any!

barbeau:
FYI, something else to consider - there are apparently `SKIPPED` stops that are supported within the OBA server now too. This effectively cancels one or more stops within a trip, but the remainder of the trip continues to run. I'm not sure what this looks like in the data yet.

aaronbrethorst (changes requested):
The `arrivals-and-departures-for-stop` API is simply filtering out canceled trips on the server side, so the iOS app will never see these passed in.

Can you:

1. close this PR.
2. Split the two fixes into different commits.
3. Submit them separately.

There's no rush to getting to getting the 'canceled' feature in if it won't be user-visible at this time, and I'd like to make sure the memory leak fix is landed sooner than later.

aaronbrethorst:
And also, regarding point 2 in the review above, please err on the side of using too many commits.

